PR #3, opened 2019-03-22
Task #3711 - jre of the properties file does not work under windows

lherschi:
- Check under windows against java.exe, instead of java, both in rust-launcher and in itweb-settings
- Unmasking the colon at properties in the rust launcher

https://icedtea.classpath.org/bugzilla/show_bug.cgi?id=3711

### DIFF
--- a/netx/net/sourceforge/jnlp/controlpanel/JVMPanel.java
+++ b/netx/net/sourceforge/jnlp/controlpanel/JVMPanel.java
@@ -54,6 +54,7 @@ import javax.swing.event.DocumentEvent;
 import javax.swing.event.DocumentListener;
 import javax.swing.text.Document;
 import net.sourceforge.jnlp.config.DeploymentConfiguration;
+import net.sourceforge.jnlp.runtime.JNLPRuntime;
 import net.sourceforge.jnlp.runtime.Translator;
 import net.sourceforge.jnlp.util.logging.OutputController;
 import net.sourceforge.jnlp.util.StreamUtils;
@@ -234,7 +235,9 @@ public class JVMPanel extends NamedBorderPanel {
             validationResult += "<span color=\"red\">" + Translator.R("CPJVMnotDir") + "</span><br />";
             latestOne = JvmValidationResult.STATE.NOT_DIR;
         }
-        File javaFile = new File(cmd + File.separator + "bin" + File.separator + "java");
+        File javaFile = new File(cmd + File.separator + "bin" + 
+                                       File.separator + "java" + 
+                                       (JNLPRuntime.isWindows() ? ".exe" : ""));
         if (javaFile.isFile()) {
             validationResult += "<span color=\"green\">" + Translator.R("CPJVMjava") + "</span><br />";
         } else {
--- a/rust-launcher/src/property.rs
+++ b/rust-launcher/src/property.rs
@@ -85,7 +85,7 @@ fn check_file_for_property(file: File, key: &str) -> Option<String> {
                     None => {}
                     Some(kvv) => {
                         if kvv.key.eq(key) {
-                            return Some(escape_unicode(kvv.value));
+                            return Some(escape_unicode(kvv.value.replace(r"\:", ":")));
                         }
                     }
                 }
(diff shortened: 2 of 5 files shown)


netx/net/sourceforge/jnlp/controlpanel/JVMPanel.java
```diff
@@ -234,7 +235,9 @@ public static JvmValidationResult validateJvm(String cmd) {
             validationResult += "<span color=\"red\">" + Translator.R("CPJVMnotDir") + "</span><br />";
             latestOne = JvmValidationResult.STATE.NOT_DIR;
         }
-        File javaFile = new File(cmd + File.separator + "bin" + File.separator + "java");
+        File javaFile = new File(cmd + File.separator + "bin" + 
+                                       File.separator + "java" + 
+                                       (JNLPRuntime.isWindows() ? ".exe" : ""));
```
judovana: Isn't this more spread? If so, maybe it can get its own function?

rust-launcher/src/property.rs
```diff
@@ -85,7 +85,7 @@ fn check_file_for_property(file: File, key: &str) -> Option<String> {
                     None => {}
                     Some(kvv) => {
                         if kvv.key.eq(key) {
-                            return Some(escape_unicode(kvv.value));
+                            return Some(escape_unicode(kvv.value.replace(r"\:", ":")));
```
judovana: Can you please explain why it is necessary?
Will it not be soved by https://github.com/rust-lang/rust/issues/27791  ? Which should be heading to stable soon.

netx/net/sourceforge/jnlp/controlpanel/JVMPanel.java
lherschi: Sorry i don't understand wath you mean with "Isn't this more spread?" I am not an English native speaker.

rust-launcher/src/property.rs
lherschi: The colon is a delimiter for Java properties. Therefore it is masked during serialization. When used in Java getProperty, it will automatically unmask. In Rust you have to take care of the unmasking yourself. How it should help escape_unicode (), does not open to me.

judovana: Colon xor equals is delimiter, depends what comes first. Isnt that?

judovana: ping?

netx/net/sourceforge/jnlp/controlpanel/JVMPanel.java
judovana: Nor am I. I suspect that java part could forget .exe  on more places. So maybe it would be good to ahve proper metho to appned .exe if necessary.

lherschi: All others I know are in the rust sources.

rust-launcher/src/property.rs
lherschi: The following line has itweb-settings written in my deplyments.properties.

deployment.jre.dir=C\:\\Program Files\\AdoptOpenJDK\\jdk-8.0.202.08\\jre

Here is the equal sign the delimiter, but java has masked the colon anyway. Without the replace, this is an invalid path, which will simply not be accepted.

lherschi: The following line has itweb-settings written in my deployments.properties.

deployment.jre.dir=C\:\\Program Files\\AdoptOpenJDK\\jdk-8.0.202.08\\jre

Here is the equal sign the delimiter, but java has masked the colon anyway. Without the replace, this is an invalid path, which will simply not be accepted.

lherschi: The following line has itweb-settings written in my deployments.properties.

`deployment.jre.dir=C\:\\Program` `Files\\AdoptOpenJDK\\jdk-8.0.202.08\\jre`

Here is the equal sign the delimiter, but java has masked the colon anyway. Without the replace, this is an invalid path, which will simply not be accepted.

judovana: Ok. gave sense now.

PR #8, opened 2019-04-01
Task #3720 - System jars (jars of the jre itself) are not recognized …

lherschi:
…as such under windows

https://icedtea.classpath.org/bugzilla/show_bug.cgi?id=3720

### DIFF
--- a/netx/net/sourceforge/jnlp/runtime/JNLPPolicy.java
+++ b/netx/net/sourceforge/jnlp/runtime/JNLPPolicy.java
@@ -47,7 +47,7 @@ public class JNLPPolicy extends Policy {
     /** the previous policy */
     private static Policy systemPolicy;
 
-    private final String jreExtDir;
+    private final URI jreExtDir;
 
     /** the system level policy for jnlps */
     private Policy systemJnlpPolicy = null;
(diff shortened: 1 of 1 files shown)


judovana:
Hi! The patch looks good. Nice catch.  The only nit is the commit message and description.
If we would be on classpath.org, I would add chenagelolg and commit message based on https://icedtea.classpath.org/bugzilla/show_bug.cgi?id=3720 . And mentioned pr3720 into News. However we are not, and new schema is being organised arround this.

I would vote for  removing Task #3720 from commit message. If nothing else, it is bug. In icedtea convention it woud be id of  PR3720.  But again, we are no longer on classapth.

Generally I would vote for dropping the "Task #3720 - " preffix, and would include https://icedtea.classpath.org/bugzilla/show_bug.cgi?id=3720 as commit message, so the content of the bug is not duplicated. WDYT?

lherschi:
From me, change the commit message according to your preferences. The task-prefix was just a suggestion from the netbeans if i link the commit to the entry from bugzilla.

judovana:
Ok. I will wait for tomorrow meeting with Adopt and Karakun,

judovana:
Still  I would like to avoid any changes to PR, the PR should be final, and directly merged with full author-ship (but that can be my personal opinion) As  maintainer of classpath ITW, I was trying to make life of other develoeprs as easy as possible,  as the patch process was not most friendly. None of above is any longer true on github.

lherschi:
No problem, just write me the desired commit message and I'll make a new pull request

judovana:
ty, then we need to cut it down to 72 chrs too:
```
Jars of the jre itself are not recognized as such under windows
see:  https://icedtea.classpath.org/bugzilla/show_bug.cgi?id=3720
```

Please update this PR if possible - it is already linked from bug and  have discussion

lherschi:
see new pull request:
https://github.com/AdoptOpenJDK/icedtea-web/pull/9

judovana:
#9 merged

PR #9, opened 2019-04-01
Jars of the jre itself are not recognized as such under windows

lherschi:
see:  https://icedtea.classpath.org/bugzilla/show_bug.cgi?id=3720


judovana:
As there are currently no  guidelines, looks ok.

PR #12, opened 2019-04-02
Packed resources are stored in cache with double file extension

lherschi:
see: https://icedtea.classpath.org/bugzilla/show_bug.cgi?id=3714


judovana:
same as PR #13  Can it be accompanied by test? If so, can this PR be updated by with commit or overwritten?

judovana:
hi Lars. Can you rebase please? There is going to happen a lot of breaking changes. I will try to merge yours PRs before  those happens

PR #13, opened 2019-04-02
Packed resources are unnecessarily cached after unpacking

lherschi:
see: https://icedtea.classpath.org/bugzilla/show_bug.cgi?id=3715

### DIFF
--- a/netx/net/sourceforge/jnlp/cache/ResourceDownloader.java
+++ b/netx/net/sourceforge/jnlp/cache/ResourceDownloader.java
@@ -357,11 +357,11 @@ public class ResourceDownloader implements Runnable {
             // return ".gz", so if we check gzip first, we would end up
             // treating a pack200 file as a jar file.
             if (packgz) {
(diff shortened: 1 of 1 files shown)


judovana:
Hi, tahnx for reopening. The code looks good, howeve, can it be accompanied by unit tests? There are several examples in current tests which are using local :  ServerLauncher s =ServerAccess.getIndependentInstance(someDirWithResource, ServerAccess.findFreePort());

Btw, I can be missing something, but you should be able to fully update PRs or even overwrite them. So the new commits are even fully overwritten, but PR is still same. It is saving quite a typing.  WDYT?

lherschi:
Sorry, I have zero experience with unit tests. Your reference to the existing examples therefore does not really help me. The only new method is markForDelete and I can not think of a useful unit-test. For the other methods where the signature has changed, I've just removed the nonsensical passing of member variables in a private method.

judovana:
Yah, was expected:)

Issue #16, opened 2019-04-02
Minimum Set of Project bootstrapping


karianna:
**🆕🐥🐶 First Timers Only**
This issue is reserved for people who never contributed to Open Source before. We know that the process of creating a pull request is the biggest barrier for new contributors. This issue is for you 💝

**👾 Description of the issue**
The following small changes / additions should be done for the repository
- [x] README.md
- [ ] CONTRIBUTING.md - including IP flow (e.g. signed commits), code quality and a *security* review.  We use GitHub PR templates for this.
- [x] LICENSE
- [ ] Tests / CI (Jenkins as a minimum, Travis or others as need be)
- [ ] Release Notes
- [ ] Clear Versioning (Semver optional but preferred)

**📋 Step by Step**
🙋 Claim this issue: Comment below.
🍴 fork the repository in github by simply clicking the 'fork' button.
⤵️ check out the forked repository
🔀 create a feature branch for the issue. We do not have any naming definition for branches.
💾 Commit your changes.
🔀 Start a Pull Request.
🏁 Done 👍 Ask in comments for a review :)
🔬 If the reviewer find some missing peaces or a problem he will start a discussion with you and describe the next steps how the problem can be solved.
💥 you did it! We will merge the fix in the master of the IcedTeaWeb project.
💐 Thanks, thanks, thanks for being part as an open source contributor for IcedTeaWeb
🤔❓ Questions

You can join our community [chat at Slack](http://adoptopenjdk.slack.com)

PR #120, opened 2019-04-04
code of conduction & styleguide

hendrikebbers:
As part of #16


hendrikebbers:
Maybe @karianna can have a look :)

judovana:
Instead of code foramtting guidelines, isnt better to introduce auto formatter setup?

hendrikebbers:
@judovana will be a next step. The guideline is not about formatting but about code pattern (like immutability & null checks)

PR #122, opened 2019-04-05
move some common tools to common module

AndreasEhret:


hendrikebbers:
ok for now but as one of the next steps we should define a new root package name (maybe `net.adoptopenjdk.icedteaweb`)

PR #125, opened 2019-04-05
move Translator to common module

AndreasEhret:
moved static config part closer to config package for now. Not perfect but breaks the dependency of Translator to DeploymentConfiguration

PR #127, opened 2019-04-05
Options definitions split

AndreasEhret:
I've mainly splitted OptionsDefinitions by extracting the options print part to separate class so that we can move OptionsDefinitions to common module. I've used a new namespace as proposed: net.adoptopenjdk.icedteaweb. This PR is based on #125


core/src/main/java/net/sourceforge/jnlp/runtime/HtmlBoot.java
```diff
@@ -47,9 +47,8 @@
 import net.sourceforge.jnlp.util.optionparser.OptionParser;
 import net.sourceforge.swing.SwingUtils;
 
-import javax.swing.JFrame;
-import java.awt.Point;
-import java.awt.Rectangle;
+import javax.swing.*;
+import java.awt.*;
```
hendrikebbers: please do not use * imports

core/src/main/java/net/sourceforge/jnlp/util/OptionsDefinitionsPrinter.java
```diff
@@ -0,0 +1,149 @@
...
+                break;
+            case TextsProvider.POLICY_EDITOR:
+                printOptions(getPolicyEditorOptions());
+                break;
+            default:
+                break;
+        }
+    }
+
+    private static void printOptions(final List<OPTIONS> options) {
```
hendrikebbers: null check for options

core/src/main/java/net/sourceforge/jnlp/util/OptionsDefinitionsPrinter.java
```diff
@@ -0,0 +1,149 @@
...
+        final List<OPTIONS> l = new ArrayList<>();
+        l.addAll(getJavaWsRuntimeOptions());
+        l.addAll(getJavaWsControlOptions());
+        //trustall is not returned by getJavaWsRuntimeOptions
+        //or getJavaWsControlOptions, as it is not desired in documentation
+        l.add(OPTIONS.TRUSTALL);
+        return l;
+    }
+
+    public static void main(String[] args) throws IOException {
```
hendrikebbers: check that args is not null and has size > 0

core/src/main/java/net/sourceforge/jnlp/runtime/HtmlBoot.java
AndreasEhret: done

core/src/main/java/net/sourceforge/jnlp/util/OptionsDefinitionsPrinter.java
```diff
@@ -124,6 +123,11 @@
     }
 
     public static void main(String[] args) throws IOException {
+        if ((args == null) || (args.length == 0)) {
+            System.out.println(String.format("Missing one of the arguments: %s | %s | %s",
```
hendrikebbers: LOG.error ;)

core/src/main/java/net/sourceforge/jnlp/util/OptionsDefinitionsPrinter.java
hendrikebbers: now you changed the functionality of the method.

I would just add Objects.requireNotNull(options); as first line of the method. if options is null a NPE would be thrown in the old code somewhere in the logic. By adding this line NPEs are much easier to find.

AndreasEhret: done

core/src/main/java/net/sourceforge/jnlp/util/OptionsDefinitionsPrinter.java
AndreasEhret: done

core/src/main/java/net/sourceforge/jnlp/util/OptionsDefinitionsPrinter.java
AndreasEhret: done

PR #132, opened 2019-04-08
Removed rhino - was applet only

judovana:


core/pom.xml
```diff
@@ -38,11 +38,6 @@
             <artifactId>mslinks</artifactId>
             <version>1.0.5</version>
         </dependency>
-        <dependency>
-            <groupId>rhino</groupId>
-            <artifactId>js</artifactId>
-            <version>1.7R1</version>
-        </dependency>
```
hendrikebbers: This will not compile. We have classes that import rhino for now. We need to remove this classes, too.

PR #133, opened 2019-04-08
Security refactoring

hendrikebbers:
Some refactoring of security classes / functionality-


core/src/main/java/net/sourceforge/jnlp/security/VariableX509TrustManager.java
```diff
@@ -70,109 +72,83 @@
 
 final public class VariableX509TrustManager {
 
-    /** TrustManagers containing trusted CAs */
-    private X509TrustManager[] caTrustManagers = null;
+    public final static String SUN_X_509 = "SunX509";
```
AndreasEhret: Consider use the const in CertificateUtils.X_509 here

PR #134, opened 2019-04-08
Logging in all classes refactored

hendrikebbers:
SLF4J API is used now for logging


FlashFreeBoard:
The console will be preserved? By this changeset, its being silenced.

hendrikebbers:
@FlashFreeBoard SLF4J can print (based on its implementation to the console). If you have slf4j-simple on the class path everything will be dumped to the console by default.

FlashFreeBoard:
also please squeeze this commit ;) (many commits before would deserve squeezing)

FlashFreeBoard:
> @FlashFreeBoard SLF4J can print (based on its implementation to the console). If you have slf4j-simple on the class path everything will be dumped to the console by default.

Nope. I'm speekign about internal console. Not about temirnal output. I guess taht will not be touched. As terminal  is remanig superpowerfull.

Itw had logging whch went to through that bottleneck, which - according to settings-  was delegated to  stdout/err and/or   internal console and/or file and/or standart java logging.

The internal conosole is relict from appelts, but it proved itself pretty useful for javaws to.  By on click, user - even on winfows - got full trace to send.

hendrikebbers:
@FlashFreeBoard ok, got it :) You mean the logging of an executed app, right?

judovana:
> @FlashFreeBoard ok, got it :) You mean the logging of an executed app, right?

I'm not sure what you mean. Sorry for bad nick. Forgot to swap.

hendrikebbers:
@judovana ok, than I'm not sure what you mean ;) In this PR I only change logging output of the WS-app/runtime to a new logger-api. This logger api can be used to do whatever you want with the logging. So we could add a WebStart-App-Logging-Frame to show it if we want.

judovana:
Yes, you had oposite approach.
Originally it was  that it went through bottleneck. and that decided where to move it. One of the ends was standart logging. So this PR is actually useless. You could just pernamently enable standart logger in bottleneck.

of course you can  go by your way - to refactor all,  and add logging backend, which will be posting to virtual console, another which will be posting to stdout/err and so on. 

Note that those were configureable from control panel via properties.  Not sure what to persist what not. Jus tletting you know what is heidden behind the lines you currently dropped.

Check the console first; It is quite cool :)

hendrikebbers:
@judovana maybe we can do a short zoom call about that topic the next days?

judovana:
sure, can be. Note that  I do not insitst on persisting of console.

hendrikebbers:
@judovana want to understand the console functionality completely before continue here and make a decision. Here a call would be awesome. Maybe later today (in 2 hours or so)?

PR #137, opened 2019-04-09
HttpUtils to common

AndreasEhret:


common/src/main/java/net/adoptopenjdk/icedteaweb/http/HttpUtils.java
```diff
@@ -34,39 +34,40 @@
...
+import org.slf4j.LoggerFactory;
 
 public class HttpUtils {
+    private final static Logger LOG = LoggerFactory.getLogger(HttpUtils.class);
 
     /**
      * Ensure a HttpURLConnection is fully read, required for correct behavior.
      * Captured IOException is consumed and printed
-     * @param c the connection to be closed silently
+     * @param c the connection to be closed silentlyboolean isHeadRequest = request.equals(
```
hendrikebbers: I think here a find&replace destroyed the doc ;)

core/src/main/java/net/sourceforge/jnlp/cache/ResourceDownloader.java
```diff
@@ -54,7 +56,7 @@ public ResourceDownloader(Resource resource, Object lock) {
         this.lock = lock;
     }
 
-    static int getUrlResponseCode(URL url, Map<String, String> requestProperties, ResourceTracker.RequestMethods requestMethod) throws IOException {
+    static int getUrlResponseCode(URL url, Map<String, String> requestProperties, HttpMethod requestMethod) throws IOException {
```
hendrikebbers: final params

core/src/main/java/net/sourceforge/jnlp/cache/ResourceDownloader.java
```diff
@@ -67,7 +69,7 @@ static int getUrlResponseCode(URL url, Map<String, String> requestProperties, Re
      * HttpURLConnection.HTTP_OK and null if not.
      * @throws IOException
      */
-    static UrlRequestResult getUrlResponseCodeWithRedirectonResult(URL url, Map<String, String> requestProperties, ResourceTracker.RequestMethods requestMethod) throws IOException {
+    static UrlRequestResult getUrlResponseCodeWithRedirectonResult(URL url, Map<String, String> requestProperties, HttpMethod requestMethod) throws IOException {
```
hendrikebbers: final params

core/src/main/java/net/sourceforge/jnlp/cache/ResourceDownloader.java
```diff
@@ -287,7 +289,7 @@ protected UrlRequestResult findBestUrl(Resource resource) {
         OutputController.getLogger().log(OutputController.Level.ERROR_DEBUG, "Finding best URL for: " + resource.getLocation() + " : " + options.toString());
         OutputController.getLogger().log(OutputController.Level.ERROR_DEBUG, "All possible urls for "
                 + resource.toString() + " : " + urls);
-        for (ResourceTracker.RequestMethods requestMethod : ResourceTracker.RequestMethods.getValidRequestMethods()) {
+        for (HttpMethod requestMethod : validRequestMethods) {
```
hendrikebbers: final

core/src/test/java/net/sourceforge/jnlp/TinyHttpdImpl.java
```diff
@@ -136,8 +135,8 @@ public void run() {
                     String request = t.nextToken();
                     String filePath = t.nextToken();
 
-                    boolean isHeadRequest = request.equals(ResourceTracker.RequestMethods.HEAD.toString());
-                    boolean isGetRequest = request.equals(ResourceTracker.RequestMethods.GET.toString());
+                    boolean isHeadRequest = request.equals(HttpMethod.HEAD.name());
```
hendrikebbers: we should avoid to use `a.equals(b)` and use `Objects.equals(a, b)` instead. By doing so no `NPE` will happen if a is null.

core/src/test/java/net/sourceforge/jnlp/TinyHttpdImpl.java
```diff
@@ -136,8 +135,8 @@ public void run() {
...
+                    boolean isGetRequest = request.equals(HttpMethod.GET.name());
```
hendrikebbers: Here is can see some problems by using enum.name():
- lower to upper case
- refactoring of enum may break this code.

Maybe we can define a String name in the enum class:

```
enum MyEnum {
A("1"), B("2");

private final String name;

public MyEnum(final String name) { this.name = name;}

public String getName() {return name;}

}
```

core/src/test/java/net/sourceforge/jnlp/cache/ResourceDownloaderTest.java
```diff
@@ -129,10 +129,10 @@ public void getUrlResponseCodeTestWorkingHeadRequest() throws Exception {
         redirectErr();
         try {
             File f = File.createTempFile(nameStub1, nameStub2);
-            int i = ResourceDownloader.getUrlResponseCode(testServer.getUrl(f.getName()), new HashMap<String, String>(), ResourceTracker.RequestMethods.HEAD);
+            int i = ResourceDownloader.getUrlResponseCode(testServer.getUrl(f.getName()), new HashMap<String, String>(), HttpMethod.HEAD);
             Assert.assertEquals(HttpURLConnection.HTTP_OK, i);
             f.delete();
-            i = ResourceDownloader.getUrlResponseCode(testServer.getUrl(f.getName()), new HashMap<String, String>(), ResourceTracker.RequestMethods.HEAD);
+            i = ResourceDownloader.getUrlResponseCode(testServer.getUrl(f.getName()), new HashMap<String, String>(), HttpMethod.HEAD);
```
hendrikebbers: Are the concrete String generic definition for the HashMap really needed? Wouldn't' diamond operator with `new HashMap()<>` work?

core/src/test/java/net/sourceforge/jnlp/TinyHttpdImpl.java
sclassen: I would not call this method getName() as it is too close to name() and the name of the method does not give a hint when to use which method.

Maybe use `request.equalsIgnoreCase(HttoMethod.GET.name())` I think it is safe to assume that the Enum HttpMethod will contain valid Http method names

common/src/main/java/net/adoptopenjdk/icedteaweb/http/HttpUtils.java
```diff
@@ -34,39 +34,40 @@
...
      */
-    public static void consumeAndCloseConnectionSilently(HttpURLConnection c) {
+    public static void consumeAndCloseConnectionSilently(final HttpURLConnection c) {
         try {
             consumeAndCloseConnection(c);
         } catch (IOException ex) {
-            OutputController.getLogger().log("Following exception: '" + ex.getMessage() + "' should be harmless, but may help in finding root cause.");
-            OutputController.getLogger().log(ex);
+            LOG.debug("Following exception: '" + ex.getMessage() + "' should be harmless, but may help in finding root cause.");
+            LOG.error("ERROR", ex);
```
hendrikebbers: This do not make sense ;) Use 1 LOG.error with the message that is in the debug (instead of "ERROR")

`LOG.error("Following exception should be harmless, but may help in finding root cause.", ex)`

common/src/main/java/net/adoptopenjdk/icedteaweb/http/HttpUtils.java
AndreasEhret: jep! thx ;-)

common/src/main/java/net/adoptopenjdk/icedteaweb/http/HttpUtils.java
AndreasEhret: agree ;), fixed

core/src/main/java/net/sourceforge/jnlp/cache/ResourceDownloader.java
AndreasEhret: done

core/src/main/java/net/sourceforge/jnlp/cache/ResourceDownloader.java
AndreasEhret: done

core/src/main/java/net/sourceforge/jnlp/cache/ResourceDownloader.java
AndreasEhret: done

core/src/test/java/net/sourceforge/jnlp/TinyHttpdImpl.java
AndreasEhret: Good point, fixed.

core/src/test/java/net/sourceforge/jnlp/cache/ResourceDownloaderTest.java
AndreasEhret: works without :) done

PR #143, opened 2019-04-09
Allow installer to be built on AdoptOpenJDK infra

charliegracie:
Backport from master

Currently the IcedTea-Web Windows installer can only be
built using an Admin account. By disabling the verification
it can be built by any user.

Signed-off-by: Charlie Gracie <charlie.gracie@gmail.com>

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -512,7 +512,7 @@ $(WIN_INSTALLER_DIR)/itw-installer.json: clean-win-installer
 win-installer: win-only-image $(WIN_INSTALLER_DIR)/itw-installer.json
 	"$(JAVA)" -jar "$(WIXGEN_JAR)" "$(DESTDIR)$(prefix)" -c $(WIN_INSTALLER_DIR)/itw-installer.json -o $(WIN_INSTALLER_DIR)/itw-installer.wxs
 	cd $(WIN_INSTALLER_DIR) && "$(WIX_TOOLSET_DIR)"/candle.exe /nologo itw-installer.wxs
-	cd $(WIN_INSTALLER_DIR) && "$(WIX_TOOLSET_DIR)"/light.exe /nologo -ext WixUIExtension itw-installer.wixobj
+	cd $(WIN_INSTALLER_DIR) && "$(WIX_TOOLSET_DIR)"/light.exe /nologo -sval -ext WixUIExtension itw-installer.wixobj
 endif
 
 # note that this is called only from windows specific target (hidden otherwise)


sclassen:
If @judovana does not review within a reasonable time frame I would go ahead and merge the 1.7 and 1.8 branches. They have no effect onto the master and we can always create new branches if we find the ones we merged are somehow incomplete

judovana:
Sorry,. I had overlooked this PR. I had backported ths patch from head to 1.8 yesterday during mass backport. Sorry and thank you for contribution (btw the issue that everybody arround do not need root to build MSI is very suspicious). Maybe we can ask alex?

akashche:
Hi, I think this fix is correct.

karianna:
@judovana Was this meant to be closed? Is it being applied elsewhere?

charliegracie:
Any update on this?  I would like to be able to run builds for the 1.8 branch and this seems to be required. Also the current tip of 1.8 now appears to be in active development for 1.8.1.  Technically I need to apply this change to a branch that only contains the code for 1.80 so that the official 1.8.0 build can be completed.  After that the more recent 1.8 changes could be applied so that 1.8.1 development can continue. 

To launch a build I need a SHA/branch/tag that represents what we want built and that contains the fix for running builds on the AdoptOpenJDK infrastructure.

So to build 1.8.0 I need something that contains the same code as the previous 1.8.0 build created from the mercurial + this fix. I am not sure if such a SHA/branch/tag exists or will exist even after this PR is merged.

charliegracie:
Any update on this?  I would like to be able to run builds for the 1.8 branch and this seems to be required. Also the current tip of 1.8 now appears to be in active development for 1.8.1.  Technically I need to apply this change to a branch that only contains the code for 1.80 so that the official 1.8.0 build can be completed.  After that the more recent 1.8 changes could be applied so that 1.8.1 development can continue. 

To launch a build I need a SHA/branch/tag that represents what we want built and that contains the fix for running builds on the AdoptOpenJDK infrastructure.

So to build 1.8.0 I need something that contains the same code as the previous 1.8.0 build created from the mercurial + this fix. I am not sure if such a SHA/branch/tag exists or will exist even after this PR is merged.

judovana:
AFAIK it is already pushed. As I wrote above, I pushed the change with mass backport from master. So it should work for you.

judovana:
https://github.com/AdoptOpenJDK/icedtea-web/commit/6549c9f00e023c5fee813c19a0241d264d4f35af  ??

PR #150, opened 2019-04-10
decouple deployment configuration from legacy and swing dependency

AndreasEhret:
- decouple deployment configuration from legacy migration (v1.4 and older structures to v1.5)
- decouple deployment configuration from swing dependency


judovana:
Unless you wish to completly remove this code, ok to merge.

AndreasEhret:
So you wouldn't mind to remove this legacy prior to 1.4 -> 1.5 directory structure migration? Then we should indeed consider to remove this code parts (MigrationUtils content) for 2.0

AndreasEhret:
I've marked the MigrationUtils as deprecated.

AndreasEhret:
As result of our confcall I will remove the legacy code parts for this PR.

PR #151, opened 2019-04-11
Adjustments to the msi-installer:

lherschi:
- change from java.bat to java.exe
 - use HKLM\Software\Classes instead of HKCR, as entries are otherwise
   made in HKCU\Software\Classes and then not apply to all users
 - register jnlp-file and jnlps-protocol


hendrikebbers:
must be reviewed by @judovana 
No idea what this does :D

Maybe merge in 1.8 would be better

judovana:
@hendrikebbers next itw is 1.9 or 2?
@lherschi  - Is it possible to obtain the 1.9.0 numbers from pom somehow? (not that it is now not possible in tip, but work in progress)

hendrikebbers:
2.0 would make sense this we change a lot.

judovana:
> 2.0 would make sense this we change a lot.

agree

judovana:
@lherschi  please update it to 2.0.0 then. TYVM!

karianna:
@judovana AFAIK This PR is against master and therefore is 2.0.x

karianna:
@judovana Is this safe to merge now?

PR #153, opened 2019-04-11
Adjustments to the msi-installer:

lherschi:
- change from java.bat to java.exe
 - use HKLM\Software\Classes instead of HKCR, as entries are otherwise
   made in HKCU\Software\Classes and then not apply to all users
 - register jnlp-file and jnlps-protocol


judovana:
Same for tip. Looks good.
Nit - 1.8.1 - those numbers can be obtained from make/configure. Worthy of try?
Not sure hwo with head.

karianna:
@judovana Is this safe to land?

judovana:
Great. TYVM!

PR #157, opened 2019-04-12
Extracted more UI components to client package

AndreasEhret:
- moved swing dependent download indicator to client package, code cleanup
- moved swing dependent FileDialogFactory to client package, code cleanup


core/src/main/java/net/adoptopenjdk/icedteaweb/client/parts/downloadindicator/DefaultDownloadIndicator.java
```diff
@@ -224,34 +218,34 @@ public void actionPerformed(ActionEvent evt) {
     static class DownloadPanel extends JPanel implements DownloadServiceListener {
         private final DownloadPanel self;
 
-        private static enum States{
-            ONE_JAR, COLLAPSED, DETAILED;
-         }
+        private enum States{
```
hendrikebbers: Can we try to externalise all enums. I do not like internal enums in classes. What do you think @AndreasEhret

AndreasEhret: In general I agree. In situations where it is solely an internal enum and of no usage outside the component, I wouldn't mind to keep it inside the component. Alternatively, if the component and its bits and pieces are put in its own package it might be ok to have the Enum in its own file. But that's just my taste...

sclassen: @AndreasEhret Fulll ACK. Keep visibility as low as possible.

hendrikebbers: As said I try to avoid inner classes/interfaces/enums but can understand your point. Let‘s go with this.
I hope that I do not automatically extract it in a future PR 😂 If that will happen: sorry 😅

PR #162, opened 2019-04-12
applet functionality removed

hendrikebbers:
This removes support for applets at several places.


judovana:
hi!

Most of ths is correct. Only ... <applet-desc> should be kept supported. Can it be? This patch is removing also tat, isnt it?

hendrikebbers:
What do you mean with "..."?

judovana:
Arg :D  md syntax consumed it! it was applet-desc element. ```<applet-desc>``` is shown as ...

hendrikebbers:
ok, I will have a look :)

hendrikebbers:
I think I need to understand better what `<applet-desc>` really does in a JNLP. So let's shift this discussion to the next days :)

judovana:
> I think I need to understand better what `<applet-desc>` really does in a JNLP. So let's shift this discussion to the next days :)

Ok.  Are you already avare of appelt desc?-) Whats the result?

judovana:
if both application-desc and applet-desc  in jnlp works,  dont hesitate to push!

judovana:
ih. plerase remove --html switch and its handling

PR #163, opened 2019-04-12
JNLP servlet module added

hendrikebbers:
Module taken as copy from https://github.com/mojohaus/webstart

### DIFF
--- a/jnlp-servlet/LICENSE.txt
+++ b/jnlp-servlet/LICENSE.txt
@@ -1,0 +1,31 @@
+Copyright (c) 2005 Sun Microsystems, Inc. All Rights Reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+
+-Redistribution of source code must retain the above copyright notice, this
+ list of conditions and the following disclaimer.
+
+-Redistribution in binary form must reproduce the above copyright notice,
+ this list of conditions and the following disclaimer in the documentation
+ and/or other materials provided with the distribution.
+
+Neither the name of Sun Microsystems, Inc. or the names of contributors may
+be used to endorse or promote products derived from this software without
+specific prior written permission.
+
+This software is provided "AS IS," without a warranty of any kind. ALL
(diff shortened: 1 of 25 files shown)


judovana:
Hmm. What is this? Why so?-)
Why copypaste it and not include it?

hendrikebbers:
@judovana hope this answer your questions :)

The new module is a copy of the JNLP-servlet-lib that was provided several years ago by SUN. 
Codehaus / Mojohaus copied the sources from SUN an added them to the https://github.com/mojohaus/webstart repo.

The code contains several useful classes for JarDiff, versioning since this is de facto the lib that everyone uses on server side to offer JNLPs (that use such features). @AndreasEhret is currently working on the version, jar-version, ... classes in this repo and we thought that it would make sense to share util classes between both sides (JNLP generation on the server - JNLP parsing on the client). Next to this the server lib is still in a beta release status and we at Karakun would love to provide a supported version of the lib. Therefore we thought it would be a good idea to add it directly to the repository.

judovana:
I see. Then the incorporation is ok, but copying? It od nothave sense. the mojohouse/webstart is in central repo. I think it is better to depend on it.  It is no obstacle to support. You can stil contribute to mojohouse directly. 
Even forking this project is still better then hardcode copy (sorry:( )

hendrikebbers:
@judovana unterstand that. But If we want to share / reuse classes in both modules (ITW and Servlet) it is much faster to a refactoring here. Later everything can be contributed to Mojohouse as a PR (with an internal dependency to ICT-common since classes are shared).

hendrikebbers:
If if Mojohouse accept the PR we can delete the module here.

judovana:
I see. I think this can not be alive without support of mojohouse. Is the PR already here?

judovana:
To lower the requirement - If people deploy  jnlp servlet from mojohous, will ITW collaborate without issues with it?
As the idea of shared logic is really god, I do not wont to kill the PR

AndreasEhret:
we should take this in for now and see how we can make best use of it.

judovana:
I keep wondering what it will be used for...
Was  mojohouse at least notified about the need for changes?

PR #171, opened 2019-04-16
Replace BASE64 Encoder and Decoder with Java8 implementation

hendrikebbers:
Until now IcedTea has `BASE64Decoder` and `BASE64Encoder` class to handle Base64 encoding. Since Java has a public API for Base64 since Java 8 we can use that one. The new class `Base64Utils` contains all functionalities of the old classes (same method signature) but is based on the public Java API.

Until now the old implementation is not removed and the new is not used in code based on a problem I would love to discuss with @judovana :

In `BASE64EncoderTest` a byte array is created that is used for tests. In the creation several newlines are added to the byte array by adding os dependend bytes:

```
 static {
        encoded.addAll(Arrays.asList(part1));
        encoded.addAll(getBASE64LineEnding());
        encoded.addAll(Arrays.asList(part2));
        encoded.addAll(getBASE64LineEnding());
        encoded.addAll(Arrays.asList(part3));
        encoded.addAll(getBASE64LineEnding());
}
```

If I use the Java 8 Base64 API all tests fail based on this new lines. If I just remove them (see `Base64UtilTests`) everything works fine. First of all I would expect that the Base64 implementation of Java8 is correct and there must be a difference / problem in the old `BASE64Decoder`. But to check that I really need to do a deep dive in this topic. Maybe you have more knowledge about this topic and the specific IcedTea implementation


sclassen:
I take a look at the issue

hendrikebbers:
@sclassen lets wait if @judovana has an answer. I do not think that you will find any solution fast. If he has no idea feel free to check both implementations :)

sclassen:
I found the reason.
The existing icedtea-web implementation adds line breaks to the base64 for pretty printing.
The Java8 implementation does not

sclassen:
From what I can tell the Base64 is only used for 2 things:
- Print out certificats in PEM format in the class `CertificateUtils`
- Read in content of files in the class `PluginBridge`

@judovana 
The javadoc of the `PluginBridge` states that it is used to bridge between applet and jnlp.
Is this class still used or will it be deprecated together with the applet support?

PR #173, opened 2019-04-16
use ico-file instead of png-file as itw-icon and use it as default icon

lherschi:
for jnlp-files on windows


judovana:
hm. This was merged without proper discussin. Both java and linuxes have issues with ICO. png shoudl be preffered way. If not, then  both png and ico should be kept,  and ico shoudl be only under some circusmatnces.

judovana:
Sorry, wrongly read PR. THe png is stil here. /me shutting up.

PR #174, opened 2019-04-16
move javaws.ico into correctly directory

lherschi:


lherschi:
This is a fix for https://github.com/AdoptOpenJDK/icedtea-web/pull/172

PR #177, opened 2019-04-17
extract StringUtils

sclassen:


judovana:
Is hamcrest related to this changeset? Maybe the commit shoudl be split into two?

sclassen:
Thanks for pointing hamcrest out.
It is indeed part of the commit, as it is used in the tests.
But I forgot to set the right scope of the dependency.

Updated the commit, please review again

PR #178, opened 2019-04-18
isolate pure locale handling functionality into LocaleUtils

AndreasEhret:
isolate pure locale handling functionality into LocaleUtils class to simplify Parser and JNLPFile

### DIFF
--- a/core/src/main/java/net/sourceforge/jnlp/util/LocaleUtils.java
+++ b/core/src/main/java/net/sourceforge/jnlp/util/LocaleUtils.java
@@ -1,0 +1,88 @@
+package net.sourceforge.jnlp.util;
+
+import java.util.Locale;
+import java.util.Objects;
+import net.adoptopenjdk.icedteaweb.xmlparser.ParseException;
+
+import static net.adoptopenjdk.icedteaweb.i18n.Translator.R;
+
+public class LocaleUtils {
+
+    public enum Match { LANG_COUNTRY_VARIANT, LANG_COUNTRY, LANG, GENERALIZED }
+
+    /**
+     * Returns a {@link Locale} from a locale string. Each locale is specified by a language identifier,
+     * possibly country identifier, and possibly a variant. The syntax is as follows:
+     * <pre>
+     * locale ::= language [ "_" country [ "_" variant ] ]
+     * </pre>
+     *
+     * @param localeString the locale string
+     * @return locale of document
+     */
+    public static Locale getLocale(final String localeString) throws ParseException {
+        if (Objects.isNull(localeString) || localeString.length() < 2) {
+            throw new ParseException(R("PIllegalLocale", localeString));
+        }
+
+        final String language = localeString.substring(0, 2);
+        final String country = (localeString.length() < 5) ? "" : localeString.substring(3, 5);
+        final String variant = (localeString.length() > 7) ? localeString.substring(6) : "";
+
+        // null is not allowed n locale but "" is
+        return new Locale(language, country, variant);
+    }
+
+    /**
+     * Returns whether a locale is matched by one of more other
+     * locales. Only the non-empty language, country, and variant
+     * codes are compared; for example, a requested locale of
+     * Locale("","","") would always return true.
+     *
+     * @param requested the requested locale
+     * @param available the available locales
+     * @param matchLevel the depth with which to match locales.
+     * @return {@code true} if {@code requested} matches any of {@code available}, or if
+     * {@code available} is empty or {@code null}.
+     * @see Locale
+     * @see Match
+     */
+    public static boolean localeMatches(Locale requested, Locale[] available, Match matchLevel) {
+
+        if (matchLevel == Match.GENERALIZED)
+            return available == null || available.length == 0;
+
+        String language = requested.getLanguage(); // "" but never null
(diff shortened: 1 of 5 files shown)


core/src/main/java/net/sourceforge/jnlp/util/LocaleUtils.java
```diff
@@ -0,0 +1,88 @@
...
+     * @param available the available locales
+     * @param matchLevel the depth with which to match locales.
+     * @return {@code true} if {@code requested} matches any of {@code available}, or if
+     * {@code available} is empty or {@code null}.
+     * @see Locale
+     * @see Match
+     */
+    public static boolean localeMatches(Locale requested, Locale[] available, Match matchLevel) {
+
+        if (matchLevel == Match.GENERALIZED)
```
sclassen: can be moved to the switch statement below as a case for GENERALIZED

core/src/main/java/net/sourceforge/jnlp/util/LocaleUtils.java
```diff
@@ -0,0 +1,88 @@
...
+    public static Locale getLocale(final String localeString) throws ParseException {
+        if (Objects.isNull(localeString) || localeString.length() < 2) {
+            throw new ParseException(R("PIllegalLocale", localeString));
+        }
+
+        final String language = localeString.substring(0, 2);
+        final String country = (localeString.length() < 5) ? "" : localeString.substring(3, 5);
+        final String variant = (localeString.length() > 7) ? localeString.substring(6) : "";
+
+        // null is not allowed n locale but "" is
```
sclassen: this comment is misleading as there cannot be a null value at this point.
Please remove the comment

core/src/main/java/net/sourceforge/jnlp/util/LocaleUtils.java
```diff
@@ -0,0 +1,88 @@
...
+     *
+     * @param requested the requested locale
+     * @param available the available locales
+     * @param matchLevel the depth with which to match locales.
+     * @return {@code true} if {@code requested} matches any of {@code available}, or if
+     * {@code available} is empty or {@code null}.
+     * @see Locale
+     * @see Match
+     */
+    public static boolean localeMatches(Locale requested, Locale[] available, Match matchLevel) {
```
sclassen: - final parameters
- null checks

core/src/main/java/net/sourceforge/jnlp/util/LocaleUtils.java
```diff
@@ -0,0 +1,88 @@
...
+            return available == null || available.length == 0;
+
+        String language = requested.getLanguage(); // "" but never null
+        String country = requested.getCountry();
+        String variant = requested.getVariant();
+
+        for (Locale locale : available) {
+            switch (matchLevel) {
+                case LANG:
+                    if (!language.isEmpty()
```
sclassen: if can be replaced by a return statement
holds for all 3 if statements within this switch block

sclassen:
Also the following change from #179 is part of this PR.
Please add it.

`PIllegalLocale=Locale attribute of information element contains illegal locale {0}.`

core/src/main/java/net/sourceforge/jnlp/util/LocaleUtils.java
AndreasEhret: does not work, as we have a unit test that explicitly defines GENERALIZED and null for available. Which leads to NPE in for loop when moving GENERALIZED to switch. As because of null for available and not empty array in case of no locales this needs bigger refactoring for all callers to ensure they handle correctly, e.g. hand in empty array and localMatches then asserts available to be non null. So keep it for now.

But I added some code as your comment pointed me to another potential problem: available=null and any other match level would end in NPE.

core/src/main/java/net/sourceforge/jnlp/util/LocaleUtils.java
AndreasEhret: done

core/src/main/java/net/sourceforge/jnlp/util/LocaleUtils.java
AndreasEhret: improved

core/src/main/java/net/sourceforge/jnlp/util/LocaleUtils.java
```diff
@@ -47,40 +49,38 @@ public static Locale getLocale(final String localeString) throws ParseException
      * @see Locale
      * @see Match
      */
-    public static boolean localeMatches(Locale requested, Locale[] available, Match matchLevel) {
+    public static boolean localeMatches(final Locale requested, Locale[] available, final Match matchLevel) {
+        Assert.requireNonNull(requested, "requested");
```
sclassen: matchLevel must not be `null` otherwise the switch statement will throw a NPE

AndreasEhret: done

core/src/main/java/net/sourceforge/jnlp/util/LocaleUtils.java
AndreasEhret: done

Issue #165, opened 2019-04-15
Windows shortcut don't handle properly filename with space


douph1:
CacheUtil.clearCache(..)
look for jnlp path from args wich is broken down to two args when jnlp file name is:
test (1).jnlp
We must probably quote arg0 (jnlp file name) inside shortcut 
WindowsDesktopEntry.createWindowsMenu()
for two lines ShellLink.createLink ( run and clearcache )
with "test (1).jnlp" and strip double quote for compare inside CacheUtil.clearCache(..)

douph1:
Run shortcut:
"C:\Program Files\IcedTeaWeb\WebStart\bin\javaws.exe" file:/C:/Users/adminup/Downloads/myjnlp (2).jnlp

Uninstall shortcut sample: 
"C:\Program Files\IcedTeaWeb\WebStart\bin\javaws.exe" -Xclearcache file:/C:/Users/adminup/Downloads/myjnlp (2).jnlp

sclassen:
resolved with #180

PR #180, opened 2019-04-18
Handle jnlp filename with space for shortcuts

douph1:
Patch for issue https://github.com/AdoptOpenJDK/icedtea-web/issues/165

### DIFF
--- a/netx/net/sourceforge/jnlp/runtime/Boot.java
+++ b/netx/net/sourceforge/jnlp/runtime/Boot.java
@@ -343,14 +343,6 @@ public final class Boot implements PrivilegedAction<Void> {
         return null;
     }
 
-    private static String stripDoubleQuote(String path) {
-        if (path.length() >= 2 && path.charAt(0) == '"' && path.charAt(path.length() - 1) == '"')
-        {
-            path = path.substring(1, path.length() - 1);
-        }
-        return path;
-    }
-
     static ParserSettings init(Map<String, List<String>> extra) {
         JNLPRuntime.setSecurityEnabled(!optionParser.hasOption(OptionsDefinitions.OPTIONS.NOSEC));
         JNLPRuntime.setOfflineForced(optionParser.hasOption(OptionsDefinitions.OPTIONS.OFFLINE));
@@ -378,7 +370,7 @@ public final class Boot implements PrivilegedAction<Void> {
             List<String> optionArgs = optionParser.getMainArgs();
             if (optionArgs.size() > 0) {
                 //clear one app 
-                CacheUtil.clearCache(stripDoubleQuote(optionArgs.get(0)), true, true);
+                CacheUtil.clearCache(optionArgs.get(0), true, true);
             } else {
                 // clear all cache
                 CacheUtil.clearCache();
--- a/netx/net/sourceforge/jnlp/runtime/Boot.java
+++ b/netx/net/sourceforge/jnlp/runtime/Boot.java
@@ -343,6 +343,14 @@ public final class Boot implements PrivilegedAction<Void> {
         return null;
     }
 
+    private static String stripDoubleQuote(String path) {
+        if (path.length() >= 2 && path.charAt(0) == '"' && path.charAt(path.length() - 1) == '"')
+        {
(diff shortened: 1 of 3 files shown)


netx/net/sourceforge/jnlp/runtime/Boot.java
```diff
@@ -343,6 +343,14 @@ private static String getMainFile() throws InvalidArgumentException {
         return null;
     }
 
+    private static String stripDoubleQuote(String path) {
```
AndreasEhret: how about put private helper methods to the end of a class (behind public methods etc.), Just my personal taste ;-)

netx/net/sourceforge/jnlp/util/WindowsDesktopEntry.java
```diff
@@ -59,7 +59,7 @@ public File getDesktopIconFile() {
     public void createShortcutOnWindowsDesktop() throws IOException {
         String path = getDesktopLnkPath();
         String JavaWsBin = XDesktopEntry.getJavaWsBin();
-        ShellLink sl = ShellLink.createLink(JavaWsBin).setCMDArgs(file.getSourceLocation().toString());
+        ShellLink sl = ShellLink.createLink(JavaWsBin).setCMDArgs('"' + file.getSourceLocation().toString() + '"');
```
AndreasEhret: const char?

netx/net/sourceforge/jnlp/util/WindowsDesktopEntry.java
```diff
@@ -90,9 +90,9 @@ public void createWindowsMenu() throws IOException {
             menuDir.mkdir();
         }
         final String JavaWsBin = XDesktopEntry.getJavaWsBin();
-        final ShellLink sl = ShellLink.createLink(JavaWsBin).setCMDArgs(file.getSourceLocation().toString());
+        final ShellLink sl = ShellLink.createLink(JavaWsBin).setCMDArgs('"' + file.getSourceLocation().toString() + '"');
```
AndreasEhret: const char?

netx/net/sourceforge/jnlp/util/WindowsDesktopEntry.java
```diff
@@ -90,9 +90,9 @@ public void createWindowsMenu() throws IOException {
...
         // setup uninstall shortcut
-        final ShellLink ul = ShellLink.createLink(JavaWsBin).setCMDArgs("-Xclearcache " + file.getFileLocation().toString());
+        final ShellLink ul = ShellLink.createLink(JavaWsBin).setCMDArgs("-Xclearcache " + '"' + file.getFileLocation().toString() + '"');
```
AndreasEhret: const char?

netx/net/sourceforge/jnlp/util/WindowsDesktopEntry.java
douph1: or const String is better ? becaus we concat with Strings

douph1: private static final String DOUBLE_QUOTE = "\"";
?

sclassen: or a helper method
```
private String quoted(URL url) {
    return String.format("\"%s\"", url.toString());
}
```

And maybe call `toExternalForm()` instead of `toString()`

douph1: jdk8 impl from toString()
>    public String toString() {
>        return toExternalForm();
>    }

sclassen: Yes, I simply find `toExternalForm()` more explicit. And I don't like to rely too much onto the implementation of toString() as for my understanding the main purpose of toString is to have a human readable representation of the object and not to meet some business requirements...

netx/net/sourceforge/jnlp/runtime/Boot.java
hendrikebbers: +1

sclassen:
thanks for the rework it looks perfect now

PR #182, opened 2019-04-18
isolated pure JVM argument checking functionality into JVMUtils

AndreasEhret:
isolated pure JVM argument checking functionality into JVMUtils class to simplify Parser

### DIFF
--- a/common/src/main/java/net/adoptopenjdk/icedteaweb/jvm/JVMUtils.java
+++ b/common/src/main/java/net/adoptopenjdk/icedteaweb/jvm/JVMUtils.java
@@ -1,0 +1,113 @@
+package net.adoptopenjdk.icedteaweb.jvm;
+
+import java.util.Arrays;
+import java.util.List;
+
+public class JVMUtils {
+    /**
+     * Check that the VM args are valid and safe.
+     *
+     * Based on
+     * http://java.sun.com/javase/6/docs/technotes/guides/javaws/developersguide/syntax.html
+     *
+     * @param vmArgs a string containing the args
+     * @throws IllegalArgumentException if the VM arguments are invalid or dangerous
+     */
+    public static void checkVMArgs(final String vmArgs) throws IllegalArgumentException {
+        if (vmArgs == null) {
+            return;
+        }
+
+        final List<String> validArguments = Arrays.asList(getValidVMArguments());
+        final String[] validStartingArguments = getValidStartingVMArguments();
+
+        final String[] arguments = vmArgs.split(" ");
+        boolean argumentIsValid;
+        for (String argument : arguments) {
+            argumentIsValid = false;
+
+            if (validArguments.contains(argument)) {
+                argumentIsValid = true;
+            } else {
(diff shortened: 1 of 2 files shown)


common/src/main/java/net/adoptopenjdk/icedteaweb/jvm/JVMUtils.java
```diff
@@ -0,0 +1,113 @@
...
+        }
+    }
+
+    /**
+     * Returns an array of valid (ie safe and supported) arguments for the JVM
+     *
+     * Based on
+     * http://java.sun.com/javase/6/docs/technotes/guides/javaws/developersguide/syntax.html
+     */
+    private static String[] getValidVMArguments() {
```
sclassen: could be converted to a constant.

Best would be `private static final Set<String> VALID_VM_ARGUEMNTS'

common/src/main/java/net/adoptopenjdk/icedteaweb/jvm/JVMUtils.java
```diff
@@ -0,0 +1,113 @@
...
+
+        final String[] arguments = vmArgs.split(" ");
+        boolean argumentIsValid;
+        for (String argument : arguments) {
+            argumentIsValid = false;
+
+            if (validArguments.contains(argument)) {
+                argumentIsValid = true;
+            } else {
+                for (String validStartingArgument : validStartingArguments) {
```
sclassen: Consider extracting the body of the else branch into a private helper method.
This would eliminate the need for the entire if statement

common/src/main/java/net/adoptopenjdk/icedteaweb/jvm/JVMUtils.java
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    /**
+     * Returns an array containing the starts of valid (ie safe and supported)
+     * arguments for the JVM
+     *
+     * Based on
+     * http://java.sun.com/javase/6/docs/technotes/guides/javaws/developersguide/syntax.html
+     */
+    private static String[] getValidStartingVMArguments() {
```
sclassen: could be converted in a constant

common/src/main/java/net/adoptopenjdk/icedteaweb/jvm/JVMUtils.java
AndreasEhret: according to zoom session: we do that in a separate PR

common/src/main/java/net/adoptopenjdk/icedteaweb/jvm/JVMUtils.java
AndreasEhret: according to zoom session: we do that in a separate PR

common/src/main/java/net/adoptopenjdk/icedteaweb/jvm/JVMUtils.java
AndreasEhret: according to zoom session: we do that in a separate PR

PR #183, opened 2019-04-18
Refactor resource downloader

sclassen:
make UrlRequestResult unmodifiable and some code formatting

### DIFF
--- a/core/src/main/java/net/sourceforge/jnlp/cache/ResourceDownloader.java
+++ b/core/src/main/java/net/sourceforge/jnlp/cache/ResourceDownloader.java
@@ -57,7 +57,7 @@ public class ResourceDownloader implements Runnable {
     }
 
     static int getUrlResponseCode(final URL url, final Map<String, String> requestProperties, final HttpMethod requestMethod) throws IOException {
-        return getUrlResponseCodeWithRedirectionResult(url, requestProperties, requestMethod).result;
+        return getUrlResponseCodeWithRedirectionResult(url, requestProperties, requestMethod).responseCode;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


core/src/main/java/net/sourceforge/jnlp/cache/ResourceDownloader.java
```diff
@@ -461,7 +461,7 @@ private void markForDelete(URL location) {
             entry.unlock();
         }
     }
-    
+
     private void writeDownloadToFile(URL downloadLocation, InputStream in) throws IOException {
         byte buf[] = new byte[1024];
```
AndreasEhret: c-style declaration

core/src/main/java/net/sourceforge/jnlp/cache/ResourceDownloader.java
```diff
@@ -155,8 +156,8 @@ private void initializeFromURL(final UrlRequestResult location) throws IOExcepti
         CacheEntry entry = new CacheEntry(resource.getLocation(), resource.getRequestVersion());
         entry.lock();
         try {
-            resource.setDownloadLocation(location.URL);
-            final URLConnection connection = ConnectionFactory.getConnectionFactory().openConnection(location.URL); // this won't change so should be okay not-synchronized
+            resource.setDownloadLocation(location.redirectUrl);
+            final URLConnection connection = ConnectionFactory.getConnectionFactory().openConnection(location.redirectUrl); // this won't change so should be okay not-synchronized
             connection.addRequestProperty("Accept-Encoding", "pack200-gzip, gzip");
 
             File localFile = CacheUtil.getCacheFile(resource.getLocation(), resource.getDownloadVersion());
```
AndreasEhret: Line 164-165 condition is always false , maybe include this existing probable bug in the refactoring/cleanup
          Long size = location.length;
            if (size == null) {

core/src/main/java/net/sourceforge/jnlp/cache/ResourceDownloader.java
```diff
@@ -295,8 +296,8 @@ protected UrlRequestResult findBestUrl(final Resource resource) {
                     Map<String, String> requestProperties = new HashMap<>();
                     requestProperties.put("Accept-Encoding", "pack200-gzip, gzip");
```
hendrikebbers: we can directly create constants for the strings

sclassen: This is only step 1 of a larger refactoring.
Extractions of constants I will do later

core/src/main/java/net/sourceforge/jnlp/cache/ResourceDownloader.java
sclassen: Yes, have not touched that method yet

core/src/main/java/net/sourceforge/jnlp/cache/ResourceDownloader.java
sclassen: done

PR #187, opened 2019-04-18
TinyHttpdImplTest.urlToFilePathTest() fails on Windows

ebourg:
Hi,

The `TinyHttpdImplTest.urlToFilePathTest()` test doesn't work on Windows because it checks unix style separators in the file paths. Here is a fix that will work everywhere.


sclassen:
Thanks for the fix

ebourg:
Thank you for merging

PR #188, opened 2019-04-18
Handle jnlp filename with space for shortcuts

sclassen:


sclassen:
I will replace the constants with "\"" during backporting

Am 18. April 2019 14:20:52 MESZ schrieb Philippe Doussot <notifications@github.com>:

hendrikebbers:
> I finally dont know if using common/src/main/java/net/adoptopenjdk/icedteaweb/IcedTeaWebConstants.java
> to define DOUBLE_QUOTE is a good idea beacause IcedTeaWebConstants is not present in 1.8 branch
We created IcedTeaWebConstants in master to define exactly such global constants that are used at several places in ITW. Sure it is not part of 1.8 and 1.7 but this is no reason for me to not use it. Otherwise we could not create new classes in the master ;)

core/src/main/java/net/sourceforge/jnlp/util/WindowsDesktopEntry.java
```diff
@@ -94,9 +97,9 @@ public void createWindowsMenu() throws IOException {
             menuDir.mkdir();
         }
         final String JavaWsBin = XDesktopEntry.getJavaWsBin();
-        final ShellLink sl = ShellLink.createLink(JavaWsBin).setCMDArgs(file.getSourceLocation().toString());
+        final ShellLink sl = ShellLink.createLink(JavaWsBin).setCMDArgs(quoted(file.getSourceLocation()));
         // setup uninstall shortcut
-        final ShellLink ul = ShellLink.createLink(JavaWsBin).setCMDArgs("-Xclearcache " + file.getFileLocation().toString());
+        final ShellLink ul = ShellLink.createLink(JavaWsBin).setCMDArgs("-Xclearcache " + quoted(file.getFileLocation()));
```
hendrikebbers: `-Xclearcache` as const

sclassen: there is much more to clean up in this class.
Will postpone to not overload this PR

sclassen:
@douph1 I removed the unquoting as you did in the 1.8 branch (#180)
Thanks for the work

Issue #100, opened 2019-04-03
Iced Tea Web OpenJDK - do not use Proxy


AndreasEhret:
This issue was imported from bugzilla with bugzilla2github
# Bugzilla Bug Referenze: [3690](https://icedtea.classpath.org/bugzilla/show_bug.cgi?id=3690)
Date: 2019-01-30 09:36:00 +0000
From: Milan &lt;<milan.krause>&gt;
To:   JiriVanek &lt;<jvanek>&gt;
CC:   axel.ahlborn, dbhole, sgehwolf, unassigned

Last updated: 2019-02-07 14:26:40 +0000

## Comment 17281
Date: 2019-01-30 09:36:19 +0000
From: Milan &lt;<milan.krause>&gt;

We installed openjdk from redhat in version 1.8.0.191-1.b12 on server 2012 r2 and 2016 and try to
                use the integrated icedtea webstart. Unfortunately, the call always tries directly access to the
                internet without using the proxy settings. Is there an extra proxy setting option for the webstart
                environment?

                Thanks in advance!
            

## Comment 17282
Date: 2019-01-30 19:46:48 +0000
From: Severin Gehwolf &lt;<sgehwolf>&gt;

Is there a itweb-settings program? If so it should launch a GUI program which has a "Network" item.
                There you should see various proxy related settings.

                Alternatively itweb-settings should be able to set this via a CLI:

                $ itweb-settings -list | grep proxy
                deployment.proxy.http.port: null
                deployment.proxy.auto.config.url: null
                deployment.proxy.https.host: null
                deployment.proxy.override.hosts: null
                deployment.proxy.socks.port: null
                deployment.proxy.socks.host: null
                deployment.proxy.bypass.local: null
                deployment.proxy.http.host: null
                deployment.proxy.https.port: null
                deployment.proxy.same: false
                deployment.proxy.ftp.port: null
                deployment.proxy.bypass.list: null
                deployment.proxy.type: 3
                deployment.proxy.ftp.host: null

                Then for example:
                $ itweb-settings -set deployment.proxy.http.port 88

                HTH
            

## Comment 17283
Date: 2019-01-31 08:58:48 +0000
From: Axel Ahlborn &lt;<axel.ahlborn>&gt;

Hello Severin,
                unfortunately we had the same problems as described in this bug. By starting a jnlp based Java
                Application from a webserver we noticed different behaviours of ITW by using a Proxy or not and when
[...]

PR #189, opened 2019-04-18
PolicyFileModel fails to parse the locked policy files on Windows

ebourg:
The `openAndParsePolicyFile()` method in `PolicyFileModel` acquires a FileLock on the policy file before parsing it. It doesn't seem to work on Windows with Java 8u191, several tests fail with the following error:

    java.io.IOException: The process cannot access the file because another process has locked a portion of the file
            at java.io.FileInputStream.readBytes(Native Method)
            at java.io.FileInputStream.read(FileInputStream.java:255)
            at sun.nio.cs.StreamDecoder.readBytes(StreamDecoder.java:284)
            at sun.nio.cs.StreamDecoder.implRead(StreamDecoder.java:326)
            at sun.nio.cs.StreamDecoder.read(StreamDecoder.java:178)
            at java.io.InputStreamReader.read(InputStreamReader.java:184)
            at java.io.BufferedReader.fill(BufferedReader.java:161)
            at java.io.BufferedReader.read(BufferedReader.java:182)
            at java.io.StreamTokenizer.read(StreamTokenizer.java:500)
            at java.io.StreamTokenizer.nextToken(StreamTokenizer.java:544)
            at sun.security.provider.PolicyParser.read(PolicyParser.java:193)
            at net.adoptopenjdk.icedteaweb.client.policyeditor.PolicyFileModel.openAndParsePolicyFile(PolicyFileModel.java:113)

This can be either fixed by using a shared lock instead of an exclusive lock (but I'm not sure about the side effects), or by parsing from the channel associated to the lock instead of creating a separate FileReader.

### DIFF
--- a/core/src/main/java/net/adoptopenjdk/icedteaweb/client/policyeditor/PolicyFileModel.java
+++ b/core/src/main/java/net/adoptopenjdk/icedteaweb/client/policyeditor/PolicyFileModel.java
@@ -36,10 +36,15 @@ exception statement from your version.
 
 package net.adoptopenjdk.icedteaweb.client.policyeditor;
 
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
 import java.io.File;
 import java.io.FileReader;
 import java.io.FileWriter;
 import java.io.IOException;
+import java.io.InputStreamReader;
+import java.io.OutputStreamWriter;
+import java.nio.channels.Channels;
 import java.nio.channels.FileLock;
 import java.util.Collection;
 import java.util.Collections;
@@ -110,7 +115,7 @@ public class PolicyFileModel {
         clearPermissions();
         final FileLock fileLock = FileUtils.getFileLock(file.getAbsolutePath(), false, true);
         try {
-            parser.read(new FileReader(file));
+            parser.read(new BufferedReader(new InputStreamReader(Channels.newInputStream(fileLock.channel()))));
             keystoreInfo = new KeystoreInfo(parser.getKeyStoreUrl(), parser.getKeyStoreType(), parser.getKeyStoreProvider(), parser.getStorePassURL());
             final Set<PolicyParser.GrantEntry> grantEntries = new HashSet<>(Collections.list(parser.grantElements()));
             synchronized (permissionsMap) {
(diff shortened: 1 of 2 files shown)


sclassen:
Thx for the PR.
Congratulation you found a rabbit hole. I opened issue #100
We will wait with merging this until we have decided how to handle encodings in the PolicyFileModel.

ebourg:
I've updated the PR to (wrongly) use the default encoding like the current implementation. I'll send another PR to address #191.

sclassen:
I want to better understand why this happens on windows and what the side effects of reading from the Channel are.
I'm going to dig out an old Windows of mine and give it a shot.
Will report back tomorrow night.

ebourg:
I guess it comes down to platform specific differences in the implementation of the file locking. It looks like on Linux the lock prevents only other processes from accessing the file, but not the JVM itself. On the other hand on Windows only the locked channel can access the file, and the same JVM is blocked if it doesn't use the channel.

Looking at the FileLock javadoc it seems accessing the locked file through the channel associated to the lock is recommended to avoid side effects:

https://docs.oracle.com/javase/8/docs/api/java/nio/channels/FileLock.html

> On some systems, closing a channel releases all locks held by the Java virtual machine on the underlying file regardless of whether the locks were acquired via that channel or via another channel open on the same file. **It is strongly recommended that, within a program, a unique channel be used to acquire all locks on any given file**.

ebourg:
Digging further, there is a similar issue in the `savePolicyFile()` method. The file is written through a FileWriter instead of reusing the locked channel. But in this case no exception is thrown, the content of the file is... erased! Reusing the channel solves this issue. I'll amend my PR and add a test covering the save method.

sclassen:
I think you are on the right track. But I didn't find time to look into the issue.
I will be offline for a week.

Sorry for the delay and thanks for the good work.

ebourg:
I've updated the PR to also fix `savePolicyFile()`.

sclassen:
Finally found time to dig into this.
Definitly an improvement. Thanks.

ebourg:
Thank you for the review.

PR #193, opened 2019-04-18
Console buttons spacing

ebourg:
There is no space between the console buttons:

2019-04-19 00_40_34-Java Console

This PR adds a 10px margin between them:

2019-04-19 00_40_00-Java Console

The layout of this window could be further improved (better alignment, margin consistency) but this is a first step. Is there any objection to use an advanced layout like [MigLayout](http://www.miglayout.com)? That would significantly simplify the code.


sclassen:
Currently the focus is not on the UI, as there is a lot of other things to get working.

Long term plans are to introduce an abstraction layer between the UI and the core functionality to allow different UIs (CLI, Swing, JavaFx).
Once we are at that stage we are certainly grateful for any help we can get.

PR #202, opened 2019-04-24
Constants externalised

hendrikebbers:
- Encodings
- System parameters


AndreasEhret:
How about the idea to make enums out of the different groups of constants (Encoding, SystemProperty, ...), similar we did this for the CommandLine options? Maybe do this in a separate refactoring PR?

PR #206, opened 2019-04-26
FileLocking refactored

hendrikebbers:
- Package moved to common module
- General refactoring based on code style guide
- Replaced JNLPRuntime.isWindows with OsUtil


core/src/main/java/net/sourceforge/jnlp/util/lockingfile/LockingReaderWriter.java
```diff
@@ -82,7 +83,7 @@ public boolean isReadOnly() {
     public void lock() {
         try {
             lockedFile.lock();
-        } catch (IOException e) {
+        } catch (final IOException e) {
```
ebourg: The final modifier isn't very useful here, it makes the code more verbose for no real benefit.

core/src/main/java/net/sourceforge/jnlp/util/lockingfile/WindowsLockedFile.java
```diff
@@ -0,0 +1,29 @@
+package net.sourceforge.jnlp.util.lockingfile;
+
+import java.io.File;
+import java.io.IOException;
+
+public class WindowsLockedFile extends LockedFile {
```
ebourg: `WindowsLockedFile` simply skips the calls to `lockProcess()`. If we are going to refactor the whole package we could simply remove `WindowsLockedFile` and change `lockProcess()` do to nothing on Windows.

core/src/main/java/net/sourceforge/jnlp/util/lockingfile/LockingReaderWriter.java
hendrikebbers: Can understand that point. In general we try to define everything "as immutable as possible". Based on this the exception must be defined as final. Otherwise some could overwrite the value:
```
} catch (final IOException e) {
   //we have a lot of thinks in this catch block going on, maybe 100 lines of code
  e = null;  // somewhere this happens
// Show we could run into bugs since must devs expect that `e` is still the given exception instance
}
```
In this given example everyone can see that this is no real problem, sure. But there are other places where this is a problem. methods params are often reallocated in ITW for example and from my point of view there is no difference between a method param or such exception. Next to this I like to have thinks like this in an "all or nothing" style but I'm free to discuss the usage of `final` for catched exceptions. Maybe @AndreasEhret and @sclassen have any other point of view. For now I would love to leave it as it is since it is as it is defined in the code-styleguide.

AndreasEhret: I would vote for go along with the "as immutable as possible" idea for now even if a little more verbose now and then.

judovana:
my minor voice; + for as immutable as possible.

PR #207, opened 2019-04-26
Refactoring of "config" package

hendrikebbers:
This PR only contains the refactoring of a part of the "config" package since the PR would become much too big otherwise.
- Validators refactored, externalised and moved in extra package
- Constants moved in single interface
- Some static contants were based open method called that could end in an exception. For this constants a thread save cache was created.
- General refactoring of several classes
- Moved several classes to net.adoptopenjdk package.


judovana:
By quick eyballing. looks good.

PR #208, opened 2019-04-26
Launchers from pom

judovana:
Hi!

this is adaptng launchers to now build. By default, they are nto build. You must use profile launchers to get them. Both shell and native ones are processed.  BOTH buuilt style was dropped, libs were narrowed.

The major usage would be:

MSLINKS_SRC=/home/blah/.m2/repository/com/github/vatbub/mslinks/1.0.5/mslinks-1.0.5.jar ITW_LIBS=DISTRIBUTION JRE=/usr/lib/jvm/java-1.8.0-openjdk mvn clean install -Plaunchers

and

ITW_LIBS=BUNDLED JRE=/usr/lib/jvm/java-1.8.0-openjdk mvn clean install -Plaunchers

All jars locations can be overwritten via variables. I have not found a better way in maven-shell connection.

This is still build only, .bats are missing (for now) and images (and thus msi) will come later.

### DIFF
--- a/launchers/build.sh
+++ b/launchers/build.sh
@@ -39,7 +39,7 @@ function publishExternalLib() {
 if [ "x$ITW_LIBS" = "xDISTRIBUTION" ] ; then
   RESOURCES_SRC_TO_DEST["$1"]="$1"
 else
-   publishInternalLib "$1" "$LIB_TARGET_DIR"
+  publishInternalLib "$1" "$LIB_TARGET_DIR"
 fi
 }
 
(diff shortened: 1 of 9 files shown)


judovana:
that escalated quickly....

judovana:
Was this read at least? I would liek to keep as close to karakun launchers as possible....

hendrikebbers:
Yes, but do not really understood the changes ;) since you are the expert in this scripts I decided that it must be correct

PR #210, opened 2019-04-29
use Charset instead of String for encoding

sclassen:


judovana:
Some of the moethods are by this changeset  no longer threowing exceptions. Worthy to remove no longer valid throws clasules/catches?

sclassen:
I fixed the throws clauses and updated the commit

judovana:
not sure if the are all (can it be actually found?). You missed few @throws in javadoc. That would kill javadoc generation with doclint. Please fix.  Looks good otherwise

PR #212, opened 2019-04-30
fix travis build

sclassen:


judovana:
Why this should be needed?

sclassen:
What do you mean with this?
The caching of the `config/icedtea-web` folder?

judovana:
Yes. Why t needs to be cached?

sclassen:
If we do not cache this folder we always get the following error:

```
Tests in error: 
  net.adoptopenjdk.icedteaweb.client.parts.dialogs.security.appletextendedsecurity.UnsignedAppletTrustConfirmationTest: ${HOME}/.config/icedtea-web/.appletTrustSettings (No such file or directory)
```

Most likely the correct way to fix this is to make sure in the above test that the file is created first.
But as a quick fix I added the caching.

I can try to fix the test once I have a stable travis build

judovana:
I disagree with this pull request. It is hiding the issue. The fix is to get rid of source of  net.adoptopenjdk.icedteaweb.client.parts.dialogs.security.appletextendedsecurity.UnsignedAppletTrustConfirmationTest: ${HOME}/.config/icedtea-web/.appletTrustSettings (No such file or directory)\

Caching this directory means also persistence of possible missconfiguration. Please revert it. 

AFAIK net.adoptopenjdk.icedteaweb.client.parts.dialogs.security.appletextendedsecurity.UnsignedAppletTrustConfirmationTest: ${HOME}/.config/icedtea-web/.appletTrustSettings (No such file or directory) is not an error, but just warning message.

Although if it started to kill the travis, maybe there was some another change which (correctly) exposed that.

The absence of appletTrustSettings is (or hsould be) harmless.  So you really had hidden an issue.,

sclassen:
@judovana I have removed the caching of the ${HOME}/.config/icedtea-web
Please have a look at the final changes for this PR

judovana:
My apologise. Thank you very much for fixing it.

PR #213, opened 2019-04-30
Refactored and enhanced CommandLine command tests

AndreasEhret:
- refactored -set command
- added new tests for -help, -info, -reset, -list, -get, -check commands
- fixed -check documentation error extra name argument
- minor refactoring: extracted missing security and rust properties as constants


core/src/main/java/net/adoptopenjdk/icedteaweb/client/commandline/CommandLine.java
```diff
@@ -78,11 +77,11 @@
     public static final int ERROR = 1;
     public static final int SUCCESS = 0;
 
-    public final String PROGRAM_NAME;
+    private final String PROGRAM_NAME;
```
sclassen: use camelCase variable name as it is not `static final`

AndreasEhret: fixed

PR #214, opened 2019-05-02
Refactored component/installer desc elements

AndreasEhret:
- Refactored component desc element
- Refactored installer desc element
- some const extraction
- improved javadoc and make reference to JSR

### DIFF
--- a/core/src/main/java/net/adoptopenjdk/icedteaweb/jnlp/element/application/ApplicationDesc.java
+++ b/core/src/main/java/net/adoptopenjdk/icedteaweb/jnlp/element/application/ApplicationDesc.java
@@ -16,9 +16,10 @@
 
 package net.adoptopenjdk.icedteaweb.jnlp.element.application;
 
+import net.adoptopenjdk.icedteaweb.jnlp.element.EntryPoint;
+
 import java.util.Arrays;
 import java.util.List;
-import net.adoptopenjdk.icedteaweb.jnlp.element.EntryPoint;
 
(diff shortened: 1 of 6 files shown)


core/src/main/java/net/adoptopenjdk/icedteaweb/jnlp/element/application/ApplicationDesc.java
```diff
@@ -31,6 +32,8 @@
  * @author <a href="mailto:jmaxwell@users.sourceforge.net">Jon A. Maxwell (JAM)</a> - initial author
  */
 public class ApplicationDesc implements EntryPoint {
+    public static final String APPLICATION_DESC_ELEMENT = "application-desc";
+    public static final String JAVAFX_DESC_ELEMENT = "javafx-desc";
```
hendrikebbers: Can we create an interface with all the *DESC* constants?

AndreasEhret: I wouldn't do that right now and see no benefit. I would prefer keeping the element/attribute constants along with its class representation instead of having them isolated in one interface class. So each descriptor defines the namespace of its element/attributes. Usually there is also only one constant for the element name. As for JavaFX, however, I cannot find any "javafx-desc" element description in the JSR spec this is kind of a special case (as ApplicationDesc has a "type" and according to spec this that can be JavaFX)...

PR #215, opened 2019-05-02
eliminated top level event package

AndreasEhret:
moved its content to specific packages as event and listener classes are only used on single places


judovana:
Interesting :) Please push.

PR #218, opened 2019-05-03
add maven wrapper

sclassen:


judovana:
Hello! You have bad luck to my bad questions:( Sorry:(
What problem does this solve? What is wrong with pure mvn clean install?

sclassen:
When running pure `mvn` the currently installed maven is used.
The maven wrapper allows to fix the maven version. So it is guaranteed that always the same maven version is used for the build. Even when a user has no maven installed :wink:

judovana:
I see. If this necessary? I mean, who will be using a source-base - javadeveloper. He see pom, he acts. Done.  I really do not see reason to add several files to fix system misconfiguration.  If exact maven is necessary, that sounds like a bug in project.
I do not wont to block this, so I would like to see an review from at least one other person.

hendrikebbers:
@judovana this pattern / idea comes from Gradle where you normally have the Gradle wrapper in a project.

judovana:
And gradle wrapper is incredibly stupid and only pointing out how fragile grade is. ( I have a lot of experience with gradle).  That is not an explanation. What is PR for, if there is disagreeing voice, the question is not answered, and it is merged anyway?
AFAIK maven do not need wrapper. If it does, then this project not yet needs it. What is it solving? Currently there were pushed N new useless copypasted files.   Jsut WHYYYYY?  The only correct step is to revert it, and explain it. If explained, then ok, push it.  But it is really not fair to push PR where was disagreement. I do not sounds myself as somehow showstopper. So please....

judovana:
Can I revert it until it is  resolved?

judovana:
ping?

PR #219, opened 2019-05-03
fix dead lock between synchronize and lock on LRU Cache

sclassen:
The main change is to synchronize on `lruHandler` before calling `lruHandler.lock()`.
Otherwise sometimes a deadlock occurs where one thread holds the monitor on `lruHandler` and another thread holds the lock.

### DIFF
--- a/core/src/main/java/net/sourceforge/jnlp/cache/CacheUtil.java
+++ b/core/src/main/java/net/sourceforge/jnlp/cache/CacheUtil.java
@@ -22,24 +22,8 @@ import net.adoptopenjdk.icedteaweb.http.CloseableConnection;
 import net.adoptopenjdk.icedteaweb.http.ConnectionFactory;
 import net.adoptopenjdk.icedteaweb.i18n.Translator;
 import net.adoptopenjdk.icedteaweb.jnlp.version.Version;
-import net.sourceforge.jnlp.config.DeploymentConfiguration;
-import net.sourceforge.jnlp.config.PathsAndFiles;
-import net.sourceforge.jnlp.runtime.ApplicationInstance;
-import net.sourceforge.jnlp.runtime.JNLPRuntime;
-import net.sourceforge.jnlp.util.FileUtils;
-import net.sourceforge.jnlp.util.PropertiesFile;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
-import javax.jnlp.DownloadServiceListener;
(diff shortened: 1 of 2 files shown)


sclassen:
@judovana can you please review this.
It looks like many changes but 99% is whitespace

The main change is that I synchronize on `lruHandler` before calling `lruHandler.lock()`.
Otherwise I sometimes end in a deadlock where one thread holds the monitor on `lruHandler` and another thread holds the lock.

judovana:
Fuuuu that is a lot of synchronizations. Although it looks good, such a change should always come in with regression test and/or unittest.

sclassen:
I agree with the testing. But it is a race condition and I can not reproduce it reliably. Only about every 10th run of the code ends in a dead lock and only on windows :(
So I don't know hpw to write a test

judovana:
hit the issue, and teh run it in the loop thousnd times?

sclassen:
And if I hit the issue it dead locks the running threads...
OK i could  run yet another thread and when the others don't terminat within n secondsI can assume a deadlock and kill the jvm...

judovana:
If the test get deadlocked, junit kills it after some timeout. No need to do it on your own.  If junit do not kills it, then hanging for ever is also good.  But as you wish...

sclassen:
@judovana 
Thx for the challenge with the test.
It made me dig deeper into the issue. At the end the failing test was easier to implement than expected :smile:

judovana:
> @judovana
> Thx for the challenge with the test.

You are welcome:)

> It made me dig deeper into the issue. At the end the failing test was easier to implement than expected smile

Indeed!  The tear down was enhanced...interestingly enough:) . So the  test  run of  PluginBridgeTest.java  file exposed the deadlock always? (is not happening on my end). If not, is it wrthy to add some loop into some tests? Or write the test directly for this case?  Moreover just asking...  TYVM!

judovana:
Possible issu with this approach may be, if the cache is to full before this class. Then the afterclass wil really take long.   Maybe to pre-clear cache in beforeclass?

sclassen:
OK, I created a new test which reproduces the deadlock for both windows and linux on my machine.
This allowed me to revert the changes to `teardown()`.

I cannot think of another way than using a timeout for detecting the deadlock.
But I increased the timeout to 10 seconds which should be more than plenty of time also for a cache with many entries

judovana:
interesting test. Thank you for all the efforts. Please push. Do you consider this backportable for 1.8?

PR #220, opened 2019-05-03
fix problems with shutdown hock of OutputController

sclassen:
`OutputController.flush()` is called in the shutdown hook of the OutputController.
If flush() has not been called before the shutdown hook is executed then the code was trying to initialize the awt toolkit.
The windows implementation of the toolkit will try to register another shutdown hook.
This fails if the shutdown is already in progress. The failure is unfortunately not handled very clean by the toolkit. As a consequence a thread is left in a `wait(0)` where it will stay forever.
This change makes sure that if flush is called during shutdown and the SwingUtils are not initialized then they will not be initialized.


judovana:
Yah, looks good. TY!

PR #224, opened 2019-05-03
Generated docs

judovana:
Added generation of docs in new environment; docs aligned with current state
    
    * As version is now exposed, bumped to 2.0
    * Doccumented death of plugin
    * no longer processing authors file
    * adapted test to it and formatter usage
    * docs propagated to images; windows get removed man o


judovana:
It is not enhancement. it was always included in  itw, now it woul be missing

PR #228, opened 2019-05-13
extract prefix match functionality to further simplify JNLPFile class

AndreasEhret:


common/src/main/java/net/adoptopenjdk/icedteaweb/StringUtils.java
```diff
@@ -67,4 +69,36 @@ public static boolean isBlank(final String str) {
...
+     * @param prefixString the prefixString string
+     * @param available the strings to test
+     * @return true if the first part of the given prefixString is a prefix for any of the strings
+     * in the specified array or the specified array is empty or null, false otherwise
+     */
+    public static boolean hasPrefixMatch(final String prefixString, final String[] available) {
+        Assert.requireNonBlank(prefixString, "prefixString");
+
+        if (available == null || available.length == 0){
+            return true;
```
sclassen: from my understanding of the javadoc this should return `false`

common/src/main/java/net/adoptopenjdk/icedteaweb/StringUtils.java
```diff
@@ -67,4 +69,36 @@ public static boolean isBlank(final String str) {
...
+        }
+
+        for (final String candidate : available) {
+            final String trimmedPrefix = prefixString.split(WHITESPACE_CHARACTER_SEQUENCE)[0];
```
sclassen: this can be moved before the loop to save some work.

common/src/main/java/net/adoptopenjdk/icedteaweb/StringUtils.java
```diff
@@ -67,4 +69,36 @@ public static boolean isBlank(final String str) {
...
+            String trimmedCandidate = null;
```
sclassen: I believe the splitting of the candidates can be omitted.
Calling `candidate.startsWith(trimmedPrefix)` should yield the same result.

But before changing this we should add some more tests for corner cases.
- Strings which start with one or multiple spaces

common/src/main/java/net/adoptopenjdk/icedteaweb/StringUtils.java
AndreasEhret: you are right, Javadoc does not explicitly say this (only in the @return text). I've improved doc for this. It was also my first reaction when I saw the code that I wanted the method ti return false here. But after studying the existing test cases it looks like this behavior was by intention.  And I didn't want to change functionality here with this refactoring. So for now I would keep it like this until we have more confidence where to check for side effects when making this method more "logic".

common/src/main/java/net/adoptopenjdk/icedteaweb/StringUtils.java
AndreasEhret: good catch, done

common/src/main/java/net/adoptopenjdk/icedteaweb/StringUtils.java
AndreasEhret: would not work, see last test I've added in prefixShouldNotBeIncludedInArrayOfStrings(). Also I would not spend to much in refactoring of the logic now, see my first comment above.

PR #229, opened 2019-05-13
backport deadlock fix from #219

sclassen:
see #219


sclassen:
To keep the diff simple I did not indent the body of the `synchronized` block.
If you want, I can easily fix this

PR #234, opened 2019-05-15
Fix tests on windows

sclassen:
This PR fixes the remaining issues we have when running the test on windows


judovana:
Thank you for doing this!

PR #237, opened 2019-05-15
install all the certificates from the pkcs12 file when importing a cl…

lherschi:
…ient certificate


sclassen:
The code looks good.
I am more concerned from a security perspective.
Adding a trusted CA is way more risky than simply adding a single certificate. With the trusted CA you implicitly add all certificates which are signed by the CA...

@lherschi maybe you could give some background why you think this change is needed.

@judovana what do you think about this change?

judovana:
hi! Shouldn't ITW be just looking into system certifcates?
It shoudl not import anything.  its local keystores are for custom certs only. Everything else goes form system.  it seems this gets broken by the changeset

lherschi:
If you install a client certificate that is not from a known trustcenter, the entire certificate chain is usually shipped with the p12 files, because you know that the ca certificate is not included in the system keystores. If you do not install the CA certificate, you must also make an exception for the client certificate in the private keystore, which then ends up in the private trustet certs store.

judovana:
hmhmh. I'm unable to find a strong reasoning against:) So ok by me. 
The messages translation are auto-translated, right? Then please skip CZ one, because CZ translator is reliable man. I can not speak for Russian one, but DE and PL are rotten anyway.

Please do your best to provide a test.  Including the dialog check. ITW have headless dialogues so you can use them.

Issue #136, opened 2019-04-08
No translation in log & exception messages


hendrikebbers:
**🆕🐥🐶 First Timers Only**
This issue is reserved for people who never contributed to Open Source before. We know that the process of creating a pull request is the biggest barrier for new contributors. This issue is for you 💝

**👾 Description of the issue**
At several places the translator is used to get a translation for a message that should be logged or used as a message for an exception. This should be removed and plain Strings should be used for logging and exceptions.

Example:
 `LOG.debug(R("RFileReadOnly"));` or `LOG.debug(Translator.R("RFileReadOnly"));` should be refactored to  `LOG.debug("File is read only");`

**📋 Step by Step**
🙋 Claim this issue: Comment below.
🍴 fork the repository in github by simply clicking the 'fork' button.
⤵️ check out the forked repository
🔀 create a feature branch for the issue. We do not have any naming definition for branches.
💾 Commit your changes.
🔀 Start a Pull Request.
🏁 Done 👍 Ask in comments for a review :)
🔬 If the reviewer find some missing peaces or a problem he will start a discussion with you and describe the next steps how the problem can be solved.
💥 you did it! We will merge the fix in the master of the IcedTeaWeb project.
💐 Thanks, thanks, thanks for being part as an open source contributor for IcedTeaWeb
🤔❓ Questions

You can join our community [chat at Slack](http://adoptopenjdk.slack.com)

aboyne:
I am interested in completing this task. I am new to open-source and am keen to get started.

hendrikebbers:
Cool! Just ask if you need help :)

aboyne:
I've forked the repository and will complete over the next few days. However I unable to add the label "In Progress" as requested by the "Step by Step" procedure. Is there something specific I need to do first?

hendrikebbers:
My fault :) Will add the label later

karianna:
I've moved it to the right stage.  @aboyne do you want to be added to the icedtea-web team?

aboyne:
Thank you @karianna. Sure, I will not be super active, but I will try to contribute where I can.

medoror:
@hendrikebbers Made a PR for this chanage

PR #238, opened 2019-05-16
Remove all translation methods from logs and exception calls

medoror:
This addresses https://github.com/AdoptOpenJDK/IcedTea-Web/issues/136


karianna:
@medoror Needs rebasing

judovana:
hi! Greate work! One nit and one issue. Nit - there are formatting changes. Please push tehm as separate commit.
Second - I heavily agree with exceptions not being localised.  However... At the end, there are user facing errors, and ITW is handling them very badly. Howver some of those exception  really go to the final dialog and are useful.  On third side, java supports (not exactly in best way) localization of exceptions. Maybe follow that?
You simply let the message id in, thats probably the worst option.

I see two approaches here -  to use your commit as it is (split in two with cosmetic changes and real changes) and rework the final dialog later
Or to rethink the final dialog first.  otherwise we end in the terrible chrome-lile dinosaur saying nothing.

What should remian (and is not included in the patch) that exception should be self explaining  to the developer. So at least fix this.
Also once you remove the keys, you should remove also  the values from properties files Already this is pointing out the isse with self-explaining  trace to developer.

Thoughts?

medoror:
@judovana Sorry for the late reply.  My plan is to wrap this up this weekend. I'll push a second commit to my branch to fix the formatting and localization issues.  Question: how should formatting issues be addressed? Is there a .editorconfig or formatter.xml file that I can use?

hendrikebbers:
@medoror until now I just used IntelliJ with n o specific configuration to auto format the code. I think that eclipse will format code more or less the same way. Only the order of imports might be different but I do not see a problem here.

medoror:
@judovana Based on your suggestions, I have pushed a second commit with the following changes: 
1) Every file I changed in the first commit was formatted with the base intellij formatter. (Suggested by @hendrikebbers )
2) Statements that throw new exceptions from catch blocks will use java localisedMessage() method within the custom message
3) Custom exception messages are formatted better with headers and colons for better readability

core/src/main/java/net/adoptopenjdk/icedteaweb/client/console/JavaConsole.java
```diff
@@ -292,9 +291,9 @@ public void actionPerformed(ActionEvent e) {
             @Override
             public void actionPerformed(ActionEvent e) {
                 printMemoryInfo();
-                LOG.info(R("CONSOLErunningFinalizers"));
+                LOG.info("CONSOLErunningFinalizers");
```
hendrikebbers: Hi @medoror 
this change is not enough since now `CONSOLErunningFinalizers` would be printed in the LOG. This is not really an information that would help since `CONSOLErunningFinalizers` is the name of the parameter that should be printed in the log. When having a look at the Messages.properties file in the repo you can find all the english translations for the parameters. In this case you can find the following line:
`CONSOLErunningFinalizers = Running finalization....
`
Based on this the code should be refactored to:
`LOG.info("Running finalization....");`

The same pattern must be used for all changed where the `R` method has been removed.

medoror: Thanks for the clarification. I was not aware about message.properties. Please see my latest commit. I pulled from the upstream and rebased my commits. Let me know if I am missing anything

sclassen:
@medoror I took to liberty to cherry-pick your commits onto a new branch based on the current master.
I then reverted most of the formatting changes. Finally I squished all of the commits and created a new PR #271.

I close this PR as it simply has too many conflicts.

PR #239, opened 2019-05-16
Refactoring download attribute

AndreasEhret:
- extracted download strategy values as enum
- extracted download element names and attributes as consts

### DIFF
--- a/core/src/main/java/net/sourceforge/jnlp/Parser.java
+++ b/core/src/main/java/net/sourceforge/jnlp/Parser.java
@@ -524,7 +524,7 @@ public final class Parser {
 
         final Node dload[] = getChildNodes(node, EXT_DOWNLOAD_ELEMENT);
         for (Node dload1 : dload) {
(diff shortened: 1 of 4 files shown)


core/src/main/java/net/sourceforge/jnlp/Parser.java
```diff
@@ -518,10 +522,10 @@ private ExtensionDesc getExtension(final Node node) throws ParseException {
 
         final ExtensionDesc ext = new ExtensionDesc(name, version, location);
 
-        final Node dload[] = getChildNodes(node, "ext-download");
+        final Node dload[] = getChildNodes(node, EXT_DOWNLOAD_ELEMENT);
         for (Node dload1 : dload) {
-            final boolean lazy = "lazy".equals(getAttribute(dload1, "download", "eager"));
-            ext.addPart(getRequiredAttribute(dload1, "ext-part", null, strict), getAttribute(dload1, "part", null), lazy);
+            final boolean lazy = LAZY.getValue().equals(getAttribute(dload1, ExtensionDesc.DOWNLOAD_ATTRIBUTE, LAZY.getValue()));
```
sclassen: here an `eager` changed to `lazy`

PR #243, opened 2019-05-20
cleanup code of XBasicService

sclassen:


judovana:
sure to go. My only nit would go against lambda. But that is the meter of personal taste. In this case, I'm for it, but they are not heavily use din ITW, and I'm not in favor of "lambdas at each cost".

Thank you very much for this. Plsease see the singl ein line review nit

core/src/main/java/net/sourceforge/jnlp/services/XBasicService.java
```diff
@@ -192,34 +187,30 @@ public boolean showDocument(URL url) {
...
+                return exec(cmd, urls);
             } else {
-                if (Desktop.isDesktopSupported()) {
-                    LOG.debug("using default browser");
-                    Desktop.getDesktop().browse(url.toURI());
-                    return true;
-                } else {
-                    LOG.debug("dont know what to do");
-                    return false;
-                }
```
judovana: There is change of behavior. Original codewas returning false, now it is throeing exception instead. How is it handled in consuming code?

sclassen: The else branch was unreachable. So in the inner if it would always take the if branch which is the one I preserved.
There code still returns a boolean and is not throwing an exception

```
if (a || !b) {
  // either a or not b is true
} else {
  // always not a and b
}

```

judovana: I overlooked the first condition. TY!

PR #252, opened 2019-05-22
Added embedded mode

judovana:
If itw is built in EMBEDDED mode, then it search for jre/jdk "where it is placed" right after java_home.
It is done by simple .., ../..
The search for libraries is then done in same way as for BUNDLED mode


karianna:
@douph1 Can you also comment?  I've added you to the team

judovana:
Well you use rust launchers in adopt's build o jdk with webstart, dont you?-)

douph1:
yes only rust exe ( no .bat files )

karianna:
@judovana can you add documentation as you suggested in the other PR?

PR #253, opened 2019-05-22
Embedded jdk

judovana:
If itw is built in EMBEDDED mode, then it search for jre/jdk "where it is placed" right after java_home.
It is done by simple .., ../..
The search for libraries is then done in same way as for BUNDLED mode


karianna:
@douph1 Can yo also comment?  I've added you to the team

douph1:
LGTM

judovana:
Thanx douph.  The documentation of BUNDLED, DISTRIBUTION and EMBEDDED should be documented. however, it is for default launchers only.  What about README.md in the launchers folder?

karianna:
> Thanx douph. The documentation of BUNDLED, DISTRIBUTION and EMBEDDED should be documented. however, it is for default launchers only. What about README.md in the launchers folder?

Sounds good to me

PR #257, opened 2019-05-23

### LINKED ISSUE #256: JNLP resources can be relative to codebase

SchwingSK:
Closes #256


SchwingSK:
Just saw that due to editing on Windows (LF vs CRLF), all files are totally changed. Will try to fix that.

PR #258, opened 2019-05-23

### LINKED ISSUE #256: JNLP resources can be relative to codebase

SchwingSK:
Closes #256

### DIFF
--- a/netx/net/sourceforge/jnlp/runtime/ManifestAttributesChecker.java
+++ b/netx/net/sourceforge/jnlp/runtime/ManifestAttributesChecker.java
@@ -386,8 +386,8 @@ public class ManifestAttributesChecker {
 
         boolean allOk = true;
         for (URL u : usedUrls) {
-            if (UrlUtils.equalsIgnoreLastSlash(u, codebase)
-                    && UrlUtils.equalsIgnoreLastSlash(u, stripDocbase(documentBase))) {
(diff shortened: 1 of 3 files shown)


judovana:
Please provide patch for master first, then backport to 1.8.
I read the code, and it looks good.
IS the method of equalsIgnoreLastSlash still use dafter your patch? If not, shoudl be dropped..
Shouldnt the conditionbe:
if ( 
(UrlUtils.equalsIgnoreLastSlash(u, codebase) && UrlUtils.equalsIgnoreLastSlash(u, stripDocbase(documentBase)))
||
(UrlUtils.urlRelativeTo(u, codebase) && UrlUtils.urlRelativeTo(u, stripDocbase(documentBase)))) { ..

? 
I iwll do some more testing tomorrow. Thank you for patch.

SchwingSK:
Patch for master provided(#260) as requested.
You're right, equalsIgnoreLastSlash had no other uses.
As I don't feel it's necessary to have the condition still use equalsIgnoreLastSlash, I removed it and integrated javadoc and tests into urlRelativeTo method.
I feel like maybe urlRelativeTo isn't the best method name (as it also covers equality, as mentioned in the javadoc), but I can't come up with anything better at the moment (urlRelativeOrEqualTo seems a bit long), so any suggestion is welcome.

PR #260, opened 2019-05-23

### LINKED ISSUE #256: JNLP resources can be relative to codebase

SchwingSK:
Closes #256

I have created class DummySecurityDelegate as it seems that neither Mockito nor Easymock is used and I didn't want to introduce dependencies. Will happily do if required.

### DIFF
--- a/core/src/main/java/net/sourceforge/jnlp/util/UrlUtils.java
+++ b/core/src/main/java/net/sourceforge/jnlp/util/UrlUtils.java
@@ -272,32 +272,6 @@ public class UrlUtils {
         return s;
     }
 
-    /**
-     * both urls are processed by sanitizeLastSlash before actual equals. So
-     * protocol://som.url/some/path/ is same as protocol://som.url/some/path. Even
-     * protocol://som.url/some/path\ is same as protocol://som.url/some/path/
-     *
-     * @param u1 first url to compare
-     * @param u2 second
-     * @return true if urls are equals no matter of trailing slash
-     */
-    public static boolean equalsIgnoreLastSlash(final URL u1, final URL u2) {
-        try {
-            if (u1 == null && u2 == null) {
-                return true;
-            }
-            if (u1 == null && u2 != null) {
-                return false;
-            }
-            if (u1 != null && u2 == null) {
-                return false;
-            }
-            return Objects.equals(sanitizeLastSlash(u1), sanitizeLastSlash(u2));
-        } catch (final MalformedURLException ex) {
-            throw new RuntimeException(ex);
-        }
-    }
-
     public static URL guessCodeBase(final JNLPFile file) {
         if (file.getCodeBase() != null) {
             return file.getCodeBase();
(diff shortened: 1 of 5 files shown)


hendrikebbers:
LGTM

core/src/main/java/net/sourceforge/jnlp/util/UrlUtils.java
```diff
@@ -341,6 +315,45 @@ public static boolean urlEquals(final URL u1, final URL u2) {
...
+     * protocol://som.url/some/path/ is same as protocol://som.url/some/path. Even
+     * protocol://som.url/some/path\ is same as protocol://som.url/some/path/
+     *
+     * This method returns false in case <code>url</code> contains parent directory notation "..".
+     * See JNLP specification version 9, 3.4: 'A relative URL cannot contain parent directory notations, such as "..". It must denote a file that is stored in a subdirectory of the codebase.'
+     * @param url the url to check
+     * @param codebaseUrl the url to check against
+     * @return true if <code>url</code> is relative to <code>codebaseUrl</code>
+     */
+    public static boolean urlRelativeTo(URL url, URL codebaseUrl) {
```
hendrikebbers: pls add final to both parameters

core/src/main/java/net/sourceforge/jnlp/util/UrlUtils.java
```diff
@@ -341,6 +315,45 @@ public static boolean urlEquals(final URL u1, final URL u2) {
...
+        if (codebaseUrl == url) {
+            return true;
+        }
+        if (codebaseUrl == null || url == null) {
+            return false;
+        }
+        try {
+            URL nu = sanitizeLastSlash(normalizeUrl(url));
```
hendrikebbers: pls make vars final

core/src/main/java/net/sourceforge/jnlp/util/UrlUtils.java
```diff
@@ -341,6 +315,45 @@ public static boolean urlEquals(final URL u1, final URL u2) {
...
+            URL nup = sanitizeLastSlash(normalizeUrl(codebaseUrl));
+            if (!getHostAndPort(nu).equals(getHostAndPort(nup))) {
+                return false;
+            }
+            if (!nu.getProtocol().equals(nup.getProtocol())) {
+                return false;
+            }
+            if (nu.getPath().contains("..")) {
```
hendrikebbers: please define a constant

core/src/main/java/net/sourceforge/jnlp/util/UrlUtils.java
```diff
@@ -341,6 +315,45 @@ public static boolean urlEquals(final URL u1, final URL u2) {
...
+                return false;
+            }
+            if (nu.getPath().startsWith(nup.getPath())) {
+                return true;
+            }
+        } catch (MalformedURLException | UnsupportedEncodingException | URISyntaxException e) {
+            LOG.error(IcedTeaWebConstants.DEFAULT_ERROR_MESSAGE, e);
```
hendrikebbers: Exception should be rethrown

sclassen: disagree on this. returning false is appropriate in this case.

hendrikebbers: @sclassen but the old implementation throws an exception.

hendrikebbers: In case of MalformedURLException

core/src/main/java/net/sourceforge/jnlp/util/UrlUtils.java
sclassen: while you are at it, could you rename
- `nu` -> `sanUrl`
- `nup` -> `sanCodebase`

core/src/main/java/net/sourceforge/jnlp/util/UrlUtils.java
sclassen: Yes, never the less I think the method should return false. If the URL is really malformed then it will trigger the `MalformedURLException` when ITW will try to download the resource. Which I find a better location to throw the exception.

core/src/main/java/net/sourceforge/jnlp/util/UrlUtils.java
SchwingSK: Done.

core/src/main/java/net/sourceforge/jnlp/util/UrlUtils.java
SchwingSK: Done.

core/src/main/java/net/sourceforge/jnlp/util/UrlUtils.java
SchwingSK: Done.

core/src/main/java/net/sourceforge/jnlp/util/UrlUtils.java
SchwingSK: I tend to side with @sclassen on this one, as MalformedURLException can be triggered by sanitizeLastSlash or normalizeUrl and in the current state these methods only throw MalformedURLException  because they use the URL constructor, but in effect it seems that they should never actually throw it as the constructed URL comes from an already valid URL (might be wrong on some edge cases in normalizeUrl method, I'm no expert).
That being said, I'll rethrow the exception if required :)

PR #18, opened 2019-04-03
merging TIP from release/icedtea-web-1.7

sclassen:

PR #263, opened 2019-05-24
import 1.7 release from mercurial

sclassen:
replacement for #18


judovana:
Compared with original mercurial 1.7 repo. Was identical.

PR #267, opened 2019-05-28
Integration test example

sclassen:


judovana:
I'm wondering, whether it is not better idea to separate build from test. Maybe each test can have its pom?

sclassen:
We looked at the existing (but currently in master not running) integration tests and came to the conclusion that for the sheer number of test it is just too much overhead and boilerplate to have a pom for every test.
We will handle ~95% of the tests with the approach lined out in this branch. For the rest we will most likely have a pom per test or something similar

PR #271, opened 2019-06-03
Remove all translation methods from logs and exception calls

sclassen:


medoror:
@sclassen sorry for the issues with conflicts and such.  This was a pretty large commit and I made many iterations over time.  Is there anything else I need to do?

sclassen:
no worries, I let some one else have a second look, and then this is good to go

Issue #217, opened 2019-05-02
Feature Request for Configurability of STDOUT Handling


martinhickson:
IcedTea-Web version: 1.8
Java version:  AdoptOpenJDK 11.0.3.7-hotspot on Windows 10

Between versions 1.8pre2 and 1.8, there was a change in the STDOUT handling of the Rust code relating to the parent process.  This has caused problems for our application as it writes to STDOUT and the buffer size when launching javaws from either the browser or Windows Explorer appears to be limited which causes the app to freeze (thread writing to STDOUT waits forever).  It works fine when javaws is launched from the command line.  Could you please add a checkbox to the IcedTea-Web settings app that would provide the 1.8pre2 behaviour?  The following commit is the relevant one:

[Commit 642aea9f140d7c14ce874e5bd9dfc99bcf14e073](https://github.com/AdoptOpenJDK/icedtea-web/commit/642aea9f140d7c14ce874e5bd9dfc99bcf14e073)

judovana:
hi! Icedtea-web controlpanel - the itweb-settings, have checkbox for log into stdouts. which is by default checked.
It results to deployment.log.stdstreams=false in deployment.properties.   Rus launchers understands that. Do yoiu mind to try? Is that an option for you?

judovana:
once unchecked,..It results to deployment.log.stdstreams=false in deployment.properties. Rust native  launchers (default in 1.8)  understands that. Do yoiu mind to try? Is that an option for you?

martinhickson:
Thank you for pointing that option out.  I have tried it and it helps with the IcedTea-Web console window in that the buttons no longer become unresponsive, however it doesn't help with our application itself since internally it is still writing to STDOUT, which I think is filling up the buffer of the parent process, whereas for 1.8pre2, it was working fine.  Perhaps the Rust launcher could redirect STDOUT to the Windows equivalent of /dev/null and this could be configurable as an additional checkbox in the settings app.

judovana:
I see.   Yes. in terminal-less output itw really should do this.  However, I'm wondering, that ITW  - on java side -  is using all streams (its own and application's ) in controlled way. s the "garbage"  is flowing from your java app, then the switch which is disbaling stdouts/err in java is not doing its job fully - it is obviously only sillenting ITW. I will double check this.

martinhickson:
Excellent, thank you.  Let me know your findings.

martinhickson:
[...]

PR #276, opened 2019-06-06
If streams are disabled, and console enabled, silent also client application

judovana:
Should be fixing the https://github.com/AdoptOpenJDK/IcedTea-Web/issues/217


karianna:
@judovana will also create a 1.8 backport

PR #279, opened 2019-06-06
cleanup code in class JavaConsole

sclassen:


core/src/main/java/net/adoptopenjdk/icedteaweb/client/console/JavaConsole.java
```diff
@@ -519,9 +480,9 @@ private void printThreadInfo() {
 
     public static void main(String[] args) {
 
-        final JavaConsole cconsole = new JavaConsole();
+        final JavaConsole console = new JavaConsole();
```
judovana: isnt this nbow in clash with global variable? it would be unnecessary warning

core/src/main/java/net/adoptopenjdk/icedteaweb/client/console/JavaConsole.java
```diff
@@ -452,47 +440,20 @@ public void run() {
...
-            LOG.debug("Available Classloaders: ");
-            Set<String> loaders = classLoaderInfoProvider.getLoaderInfo().keySet();
-            for (String loader : loaders) {
-                LOG.debug(loader + "\n"
-                        + "  codebase = "
-                        + classLoaderInfoProvider.getLoaderInfo().get(loader));
-            }
-            LOG.debug(" ----");
-        }
+        LOG.debug("No Classloader info exists in system");
```
judovana: Ugh. Why so?  Class laoders info are part of proprieatary javaws. It should be fixed rather then removed.

core/src/main/java/net/adoptopenjdk/icedteaweb/client/console/JavaConsole.java
```diff
@@ -452,47 +440,20 @@ public void run() {
         });
     }
 
-    public void hideConsoleLater() {
```
judovana: This is really unused?

core/src/main/java/net/adoptopenjdk/icedteaweb/client/console/JavaConsole.java
sclassen: no, there is no field in this class which is called `console`

core/src/main/java/net/adoptopenjdk/icedteaweb/client/console/JavaConsole.java
sclassen: At least the compiler does not complain about a missing method and a search over all code does not find any other hit for the string `hideConsoleLater`

core/src/main/java/net/adoptopenjdk/icedteaweb/client/console/JavaConsole.java
sclassen: Fixing is non trivial as there is currently no implementation of the interface `ClassLoaderInfoProvider`. I can open an issue about the missing implementation.

judovana:
please go on and merge.

PR #282, opened 2019-06-08
BUGFix: handle spaces in path names

vasvir:
Trying to run IcedTea-Web in MacOSX (El Capitan) fails due to space in path.
I was trying to install it under /Library/Application Support/icedtea-web-image.
With this patch I was able to run my webstart applications.
The general class of fix is that in shell always enclose in double quotes variable expansions like a="$b".
The patch fixes PORTABLE_ITW_HOME as it should.

The patch doesn't change BINARY_LOCATION (next line) when it should.
Basically BINARY_LOCATION shouldn't work but it does and so I left it unchanged.
The following attached small test demonstrates this behavior.

and here is its output if you don't feel performing the test yourself. The test is on linux but it has the same behavior in MacOSX (El Capitan)

[test.log](https://github.com/AdoptOpenJDK/IcedTea-Web/files/3268316/test.log)
[javaws.sh.txt](https://github.com/AdoptOpenJDK/IcedTea-Web/files/3268317/javaws.sh.txt)


karianna:
@vasvir Do you need this for the 1.8.x line of ITW?  If so can you also submit a backport PR to the 1.8.x branch, thanks!

PR #287, opened 2019-06-12
Set the environment for the tests so that all tests can be passed in …

lherschi:
…non-English language environments, not utf-8-encoding and with the splash screen disabled.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -64,6 +64,12 @@
                     <groupId>org.apache.maven.plugins</groupId>
                     <artifactId>maven-surefire-plugin</artifactId>
                     <version>${maven-surefire-plugin.version}</version>
+                    <configuration>
+                        <argLine>-Dfile.encoding=${project.build.sourceEncoding} -Duser.language=en</argLine>
+                        <environmentVariables>
+                            <ICEDTEA_WEB_SPLASH>DEFAULT</ICEDTEA_WEB_SPLASH>
+                        </environmentVariables>
+                    </configuration>
                 </plugin>
 
                 <plugin>


judovana:
Hi Larsch! 

This is interesting patch. I'm wondering if it should be the default. Maybe it should be some custom profile or similarly?

sclassen:
I like the idea of having the build work on all OS the same way.
I just wonder how we could test behavior of ITW when it is started on windows where the charset is not UTF-8

lherschi:
In fact, all tests are successful only if the build environment has the patching features. This is not limited to Windows, but also exists in Linux. If explicitly setting the build environment is not desired, the tests that fail will need to be customized. However, this will be difficult for the tests that depend on the charset, since there are innumerable charsets.

lherschi:
Die folgenden Tests scheitern, wenn die Umgebung nicht passt.

non utf8 environment:

TeeOutputStreamTest.testPrint()
TeeOutputStreamTest.testWriteByteArrayString()
TeeOutputStreamTest.testWriteByteArrayStringInUtf8()

non english environment:

HelpCommandTest.testHelpCommand()
InfoCommandTest.testInfoCommand()
SetCommandTest.testSetCommandDisplaysWarningOnUnknownProperty()
SetCommandTest.testSetCommandWithDuplicateKeyValue()
SetCommandTest.testSetCommandWithIdenticalKeyAndValue()
SetCommandTest.testSetCommandWithKeyHavingSpace()
SetCommandTest.testSetCommandWithValueHavingSpace()

disabled Splashscreen:

ErrorSplashScreenTest.ErrorSplashScreenTest()

PR #290, opened 2019-06-15
Path URI for with space throw java.net.URISyntaxException

Budlee:


karianna:
@Budlee if this is for 1.8.x you'll need to target that branch and/or create a separate backport PR

judovana:
looks good. For 1.8 and specific PR will need to be created. This is indeed for 2.x only

PR #293, opened 2019-06-18
Fix hang in windows when javaws is launched outside the console.

lherschi:
See issue 273.


judovana:
Hi Larsch, can this be incorrect?
How ITW internal console behaves in this scenario? Will it still get correct  output from custom app?

lherschi:
The console behaves just like before the change. I have full output on both consoles.

judovana:
Thank you. Will merge. Do you mind to backport to 1.8 please?

PR #294, opened 2019-06-19
fix(JNLPFile): location/sourceLocation confusion

### LINKED ISSUE #265: Unnecessary check of JNLP file sourceLocation
During launch, the JNLP file sourceLocation is added to the URLs to be checked against the codebase URL. If it isn't relative to the codebase (provided issue #256 is resolved), then a warning dialog is displayed:
image

In particular, this happens (as shown above) when the jnlp file is downloaded via the browser then passed to the javaws client.
This behavior is not the same as in Oracle Web Start, which did not show any warning.
This looks like a bug, as the JNLP specification, section 6.6 goes like this:
"Application descriptors, i.e., JNLP files, are handled specially, since they are not necessarily downloaded by the JNLPClient itself. Often, they will be downloaded by a Web browser or by other means."

SchwingSK:
Closes #265

### DIFF
--- a/core/src/main/java/net/adoptopenjdk/icedteaweb/client/parts/dialogs/security/appletextendedsecurity/UnsignedAppletTrustConfirmation.java
+++ b/core/src/main/java/net/adoptopenjdk/icedteaweb/client/parts/dialogs/security/appletextendedsecurity/UnsignedAppletTrustConfirmation.java
@@ -114,8 +114,9 @@ public class UnsignedAppletTrustConfirmation {
     }
 
     private static UnsignedAppletActionEntry getMatchingItem(UnsignedAppletActionStorage actionStorage, JNLPFile file, Class<? extends RememberableDialog> id) {
+        URL location = (file.getSourceLocation() != null) ? file.getSourceLocation() : file.getFileLocation();
         return actionStorage.getMatchingItem(
-                UrlUtils.normalizeUrlAndStripParams(file.getSourceLocation(), true /* encode local files */).toString(),
+                UrlUtils.normalizeUrlAndStripParams(location, true /* encode local files */).toString(),
                 UrlUtils.normalizeUrlAndStripParams(file.getNotNullProbableCodeBase(), true /* encode local files */).toString(),
                 toRelativePaths(getJars(file), file.getNotNullProbableCodeBase().toExternalForm()), id);
     }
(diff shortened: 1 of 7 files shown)


SchwingSK:
Sorry about breaking integration tests, but as these didn't pass before my commit (I'm on Windows, don't know if they are supposed to work on this OS), I kind of hoped that they wouldn't be affected.
Will try to investigate (and I'm open to suggestions on how to make these run on Windows)

judovana:
Hi Thierry! What is reason for this change? I belive it is some unhappy fallback for some strange cornercase.

SchwingSK:
Hi judovana,
This is going to be a bit long :)
The case is the following:
our application needs to contact Web Services (located on the server) from the JavaWebStart client.
In most cases the server (containing the JNLP file and the WS) relies on some SSO for authentification a autorizations.

Most HTTP SSO implementations rely on a reverse proxy that intercepts requests that are not authentified, redirect them to a login page, then send a authentication token in a cookie to the browser & redirect to the originally requested page.

Each time a request is made, the reverseProxy checks for this cookie. If it's not there, then the request is redirected to the login page.

Problem is, once our JavaWebStart (JWS) application is launched, we have no way to get this cookie (well we COULD but that would mean that the user has to login again, defeating the whole **Single** Sign On thingy. So we need to get this cookie to the JWS app. 

One way to do this is to have the server generate the JNLP dynamically and include the cookie as a launch parameter. Easy. BUT for this to work, we mustn't have the href attribute on the JNLP tag, otherwise the JWS client (here IcedTea Web, but Oracle JWS behaves similarly, and according to specification) woud try to download the JNLP file again from the server, and would get stopped by the SSO reverse proxy (no cookie...)

Oh, I forgot: this happens because web browsers first save the JNLP file in a temp folder, and then launch javaws with this file as a parameter.

The bug here was that JNLPFile confused SourceLocation (which is supposed to be the original source where the JNLP comes from) and FileLocation (which is the location of the local copy). When the href attribute is null, there's no way to infer the sourceLocation, and so sourceLocation should stay null.

Hope that clarifies things a bit... Don't hesitate to ask questions, it's quite tricky, even if the fix is simple.

SchwingSK:
Fixed integration tests. There was another place in the code that relied on the sourceLocation/fileLocation confusion. I'll have to track all usages and see if there aren't any more.

Issue #168, opened 2019-04-15
future of javaws.jar


judovana:
With the ITW now being split into into several modules, I'm wondering if assembled javaws.jar will be  part of the build. If so, then I would like to suggest optionally to no assembling external dependencies (like tagsoup or slf* or mslinks) 

I'm currently adapting the rust launchers (very slowly, time is again the enemy) to new structure, and had stumbled over the many jars from mvn  install being not convenient, and  in same, with wish to keep distributin build to use system jars of dependencies.

hendrikebbers:
I think it is a good idea to have the modules and create separate jars out of it but I‘m fine with providing an assembled jar in addition. Does this has a high priority or can we wait until the big refactoring is done?

judovana:
This have absolutely no priority, but will become important in time of your /our first release, as final jars will be hard to change yesterday. And itw should ship the same jars on all platforms/releases.
javaws.jar means comaptibility with oracle javaws. And as bugs "worsk with oracel, not on iTW" are most stupid ones, I would like to avid that. (except that fact, I 100% agree with having jar per module!!!)
I vote for javaws.jar being the single jar of ITW, but I will follow Yours release schema in linux  distributons  and "my" upstream release (if there will be some).  So it would be definitely good to agree on the approach.

The side effect of this issue is that assembly plugin do  not like to to not include external dependencies, so to maintain that, can be some unnecessary burden.  Thus openig this for discussion and other expereince:) TY!

judovana:
> I think it is a good idea to have the modules and create separate jars out of it but I‘m fine with providing an assembled jar in addition. Does this has a high priority or can we wait until the big refactoring is done?

@hendrikebbers  According to this, karakun have no intentions to provide javaws.jar.  I had adapted the default launchers to not expect that.  Thanx! (please reopen if this change)

PR #295, opened 2019-06-20
build a single javaws.jar with all dependencies included

sclassen:
To simplify the building of launchers and installers a "fat jar" of the entire ITW with all its dependencies is created.

@judovana 
I tried to understand how the shell and rust launchers are created in order to propose to only depend on the new fat jar. I think this would simplify the many things.
The fat jar includes both the compiled sources of ITW and all the class files of the maven dependencies.

Unfortunately I failed to understanding the different modes in which the launchers can be built. Especially `DISTRIBUTION` and `EMBEDDED`. Never the less I feel this change has the potential to simplify the launchers


judovana:
Hello! Thanx for great commit! See  https://github.com/AdoptOpenJDK/IcedTea-Web/issues/168 if you wish.

I already gave up on fat jar. it can introduce maybe more issues the it can fix. It will lead to big divergence between embedded/portable and distribution builds.  it will lead also to great divergence of the launchers themselves. It can lead to error caused by assembly itself (some lost or overwritten class) which are hard to catch.

Except good effort to simplyfy launchers, is there any other benefit?

distribution build is buid, which search for libraries in system. So eg fedora package of ite will have only itw's jars, and all deps will flow from system packages.
Portable is build, when jars in  ./libs have priority over system jars
embedded is like portable, only instead of system java it searches for  jave in ".."

sclassen:
would it be of any use if we provide another fat jar which only includes ITW classes??

judovana:
A bit of course.  My only main reason for yes, is that oracle provides javaws.jar and people are used for it. I think it is better then one fat jar with all deps. Still the benefit is sporadical. So I'm hesitating with full yes, and am hesitating on exacly 50/50. The jar with itw classes only is definitley better then jar with alldeps.

The possible issue, that something may slip/change in fat jar in difference agaisnt small jars is stillthere. What will karakun's launcher use? If you go with fat jar,then  it will case levrage to 51:49 FOR javaws.jar:) If not then I still remian usure.

sclassen:
We already build a fat jar as part of the build job for open web start. It is just a little unsafe as we modify the pom after downloading to add the javaws module to the build. Therefore I would prefer to have the javaw artifact in the main ITW repo.

Also I thought you may profit from this. Especially since we experience issues with the JNLPSecurityManager and the JNLPPolicy since we split up ITW into smaller jars.
Reason might be that JNLPPolicy only treats its own code source plus system as `all-priveleges` if ITW is run in secure mode

judovana:
Super-fat jar  with all deps, or itw only?
I gave it several thougts over night, and I'm deffinitly +1 for itw one.
I stil hesitate with all super fat.

Your point with security manager is really good.   But that evokes super fat, and not iw-fat only:(

Thanx again!

sclassen:
I added now a module called `itw` which builds itw-VERSION.jar which contains all ITW classes but none of the dependencies

judovana:
In this fashion, can  the  javaws module build itw-version-alldeps.jar instead of javaws.jar?
From my point of view, javaws.jar is more the thsi new itw module.  Can maybe jaavws module be renamed to itw-alldeps?

I know itis nitpicking, but the naming is now confusing. To rename the desired jar by integration compose creation should be the way. ( you will be renaming itw-alldeps, I will be renaming ite :) )

Again, thank you for doing this. it will simplyfy the launchers a lot.


You have not anwered my last question - How  the itw-only fat jar may be hurt in   the mener of JNLPSecurityManager you mentioned above?

The deps must go to bootclassapth in all cases. And it is not exactly suitable with jdk11 and up/

sclassen:
Sorry I didn't spot your question.
I still don't fully understand how SecurityManager and Policy work. From my current understanding I would expect that ITW 2.0 is not able to create desktop shortcuts and menu entries if run in secure mode. But I never tried to see what happens...

sclassen:
Regarding the renaming, I will create the following:
- itw-no-depts-VERSION.jar
- itw-all-depts-VERSION.jar

judovana:
> Regarding the renaming, I will create the following:
> 
>     * itw-no-depts-VERSION.jar
> 
>     * itw-all-depts-VERSION.jar

+1.

ty. Feel free to merge afterwards. Future  will see how it fit to final images.

sclassen:
the new jars are created in the maven module named `artifacts`

PR #298, opened 2019-06-21
support handling of white space in PolicyURI when on file system

sclassen:
Backport of #290


sclassen:
Did some testing on linux, but not much.

judovana:
nvm. it is better then it was. Feel free to push.

PR #306, opened 2019-06-26
unbreak build of launchers since merge of PR #266

lherschi:
fix for build on windows
remove unnecessary hyphen at the end of tail (breaks build on freebsd)

### DIFF
--- a/launchers/build.sh
+++ b/launchers/build.sh
@@ -5,6 +5,14 @@ set -x
 set -e
 set -o pipefail
 
+isWindows() {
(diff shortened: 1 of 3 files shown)


judovana:
Thank you for PR. Two nits. Maybe use version sort  for sorting? Amybe muve is windows to utils? Tahnx!

lherschi:
I already need isWindows in build.sh because it's already used for the value of SCRIPT_DIR. $SCRIPT_DIR is then used at the source command of utils.sh

PR #308, opened 2019-06-28
Always add system property "deployment.javaws"="IcedTea-Web"

pbi-qfs:
This makes it possible for the started application to detect the execution context.

Based on the presence of the system property, multi-deployment applications can adapt to the WebStart running context, and test frameworks can use special optimizations for WebStart runs.

### DIFF
--- a/core/src/main/java/net/sourceforge/jnlp/runtime/ApplicationInstance.java
+++ b/core/src/main/java/net/sourceforge/jnlp/runtime/ApplicationInstance.java
@@ -57,6 +57,9 @@ import java.security.ProtectionDomain;
 public class ApplicationInstance {
 
     private final static Logger LOG = LoggerFactory.getLogger(ApplicationInstance.class);
+    
+    public final static String DEPLOYMENT_SYSPROP = "deployment.javaws";
+    public final static String DEPLOYMENT_SYSPROP_VALUE = "IcedTea-Web";
 
     // todo: should attempt to unload the environment variables
     // installed by the application.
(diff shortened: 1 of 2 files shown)


judovana:
hello! Thanx for PR.

What is expected workflow with deployment.javaws ?
What is oracle javaws setting as value here?
MAybe we should lie and se tit to something similar, eg with suffix? I can imageine apps doing "if value(deployment.javaws).contains(oracle javaws)) then...
At least I have this experience from misusing java version strings and proeprties.

pbi-qfs:
In some situations, it is interesting to detect, if the application was started via WebStart, or not. For Oracle's JavaWs, we currently use in our agent:
```java
isWebStart = System.getProperty("jnlpx.home") != null;
```
but IcedTea-Web (and thus the tested OpenWebStart) does not provide such a property. In addition, since IcedTeaWeb behaves differently than Oracle's javaws with respect to the DispatchThread usage, it is interesting to differentiate in between the classic javaws and the modern open-source implementations.

Sam1970:
Hi @jodovana,

that is default JNLP Arguments. Here is Oracle Source File:
https://github.com/barchart/barchart-oracle-study/blob/master/oracle-jdk-7.21-javaws/src/main/java/com/sun/javaws/JnlpxArgs.java

You can find infornations about oracle Java WebStart too (Source Files).

Regards

AndreasEhret:
Hi @judovana,

what do you mean by suffix, something like "deployment.javaws.provider" or "deployment.javaws.launcher"?

I think we should go for a new property as proposed by @pbi-qfs, testing presence of system property jnlpx.home looks not very intuitive to me.

hendrikebbers:
LGTM
As far as I can say this only introduces a system variable that is always set with a fixed value. Since Oracle WS does the same and people already use it (if I understood this correct) we should just add it :)

PR #310, opened 2019-07-01
simplify fetching JNLP file location from command line argument

AndreasEhret:
- simplify fetching JNLP file location from command line argument
- be more tolerant for unknown extra arguments by ignoring them when fetching JNLP file location instead of throwing an InvalidArgumentException

### DIFF
--- a/core/src/main/java/net/sourceforge/jnlp/runtime/Boot.java
+++ b/core/src/main/java/net/sourceforge/jnlp/runtime/Boot.java
@@ -51,6 +51,7 @@ import java.util.List;
 import java.util.Map;
 
 import static java.nio.charset.StandardCharsets.UTF_8;
+import static java.util.Objects.isNull;
 import static net.adoptopenjdk.icedteaweb.IcedTeaWebConstants.JAVAWS;
 import static net.adoptopenjdk.icedteaweb.i18n.Translator.R;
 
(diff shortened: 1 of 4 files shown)


core/src/main/java/net/sourceforge/jnlp/runtime/Boot.java
```diff
@@ -328,24 +321,23 @@ static URL getFileLocation() {
...
+     * @return the file location or null if no file location can be found in the command line arguments.
      */
-    private static String getMainFile() throws InvalidArgumentException {
-        if (optionParser.getMainArgs().size() > 1
-                || (optionParser.mainArgExists() && optionParser.hasOption(CommandLineOptions.JNLP))
-                || (optionParser.mainArgExists() && optionParser.hasOption(CommandLineOptions.HTML))
-                || (optionParser.hasOption(CommandLineOptions.JNLP) && optionParser.hasOption(CommandLineOptions.HTML))) {
-            throw new InvalidArgumentException(optionParser.getMainArgs().toString());
-        } else if (optionParser.hasOption(CommandLineOptions.JNLP)) {
+    private static String getJnlpFileLocationFromCommandLineArguments() {
```
hendrikebbers: instead of accessing the static optionParser we should add it as a param to the method - different name than 'optionParser' ;) - and work internally only with the param. By doing so we can easily test this methods with some unit tests to check that the functionality is the same as before.

AndreasEhret: I thought about it already but wanted keep the refactoring small. But will have a look to improve to be able to write some unit tests

AndreasEhret: Refactoring done to improve testability.

hendrikebbers:
LGTM

PR #312, opened 2019-07-03
implement parsing for progress-class attributes

### LINKED ISSUE #311: progress-class attributes not parsed
progress-class attributes not parsed for application-desc, applet-desc and installer-desc as required since JSR spec 6.0.18.

The (optional) progress-class attribute can be used to indicate that the class of this name implements the javax.jnlp.DownloadServiceListener interface. This class may be loaded first and used to indicate the progress of other resources being downloaded and verified.

AndreasEhret:
implement parsing for progress-class attributes of AppletDesc, ApplicationDesc and InstallerDesc

fixes #311

### DIFF
--- a/core/src/main/java/net/adoptopenjdk/icedteaweb/jnlp/element/application/AppletDesc.java
+++ b/core/src/main/java/net/adoptopenjdk/icedteaweb/jnlp/element/application/AppletDesc.java
@@ -44,6 +44,7 @@ public class AppletDesc implements EntryPoint {
     public static final String NAME_ATTRIBUTE = "name";
     public static final String WIDTH_ATTRIBUTE = "width";
     public static final String HEIGHT_ATTRIBUTE = "height";
+    public static final String PROGRESS_CLASS_ATTRIBUTE = "progress-class";
 
     /** Name of the Applet. This is available to the Applet through the AppletContext. */
     private final String name;
@@ -56,7 +57,7 @@ public class AppletDesc implements EntryPoint {
 
     /**
      * The name of a class containing an implementation of the {@link javax.jnlp.DownloadServiceListener}
-     * interface of applications.
+     * interface of applications. May be used to indicate download progress.
      */
     private final String progressClass;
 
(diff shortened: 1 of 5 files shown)


core/src/main/java/net/adoptopenjdk/icedteaweb/jnlp/element/application/AppletDesc.java
```diff
@@ -44,6 +44,7 @@
     public static final String NAME_ATTRIBUTE = "name";
     public static final String WIDTH_ATTRIBUTE = "width";
     public static final String HEIGHT_ATTRIBUTE = "height";
+    public static final String PROGRESS_CLASS_ATTRIBUTE = "progress-class";
```
hendrikebbers: Is this used in the code?

core/src/main/java/net/sourceforge/jnlp/Parser.java
```diff
@@ -831,12 +831,11 @@ public EntryPoint getEntryPointDesc(final Node parent) throws ParseException {
...
      *
      * @throws ParseException if the JNLP file is invalid
-     *
-     * TODO: parse and set {@link AppletDesc#getProgressClass()}
      */
-    private AppletDesc getApplet(final Node node) throws ParseException {
+    AppletDesc getAppletDesc(final Node node) throws ParseException {
         final String name = getRequiredAttribute(node, AppletDesc.NAME_ATTRIBUTE, R("PUnknownApplet"), strict);
         final String main = getMainClass(node, true);
+        final String progressClass = getAttribute(node, ApplicationDesc.PROGRESS_CLASS_ATTRIBUTE, null);
```
hendrikebbers: I assume this should be AppletDesc.PROGRESS_CLASS_ATTRIBUTE

core/src/main/java/net/adoptopenjdk/icedteaweb/jnlp/element/application/AppletDesc.java
AndreasEhret: Not now, this PR is about implement proper parsing of progress-class attribute. Making use of it will be a separate PR.

core/src/main/java/net/sourceforge/jnlp/Parser.java
AndreasEhret: Thanks for seeing this, fixed.

PR #316, opened 2019-07-09
Add os and arch parsing of information desc

AndreasEhret:


core/src/main/java/net/adoptopenjdk/icedteaweb/jnlp/element/information/InformationDesc.java
```diff
@@ -234,9 +294,24 @@ public boolean isSharingAllowed() {
      * @return the associations specified in the JNLP file
      */
     public AssociationDesc[] getAssociations() {
-        List<Object> associations = getItems(ASSOCIATION_ELEMENT);
+        List<AssociationDesc> associations = (List) getItems(ASSOCIATION_ELEMENT);
```
hendrikebbers: Null check should be added since we use it later

core/src/main/java/net/sourceforge/jnlp/Parser.java
```diff
@@ -630,7 +630,9 @@ private InformationDesc getInformationDesc(final Node node) throws ParseExceptio
         final List<String> descriptionsUsed = new ArrayList<>();
 
         // create information
-        InformationDesc informationDesc = new InformationDesc(getLocales(node), strict);
+        final String os = getAttribute(node, InformationDesc.OS_ATTRIBUTE, null);
+        final String arch = getAttribute(node, InformationDesc.ARCH_ATTRIBUTE, null);
+        InformationDesc informationDesc = new InformationDesc(getLocales(node), os, arch, strict);
```
hendrikebbers: final

core/src/main/java/net/adoptopenjdk/icedteaweb/jnlp/element/information/InformationDesc.java
AndreasEhret: association will not be null, always returns an empty list

core/src/main/java/net/sourceforge/jnlp/Parser.java
AndreasEhret: done

PR #319, opened 2019-07-12
Extract JVM launcher interface

AndreasEhret:
- to enable changing the JVM version or passing different/extra arguments


core/src/main/java/net/adoptopenjdk/icedteaweb/impl/ApplicationImpl.java
```diff
@@ -2,29 +2,27 @@
...
 
+import java.net.URL;
 import java.util.concurrent.CompletableFuture;
 
 public class ApplicationImpl implements Application {
 
-    @Deprecated
-    private final JNLPFile file;
-
+    private final URL jnlpFileLocation;
```
sclassen: What do we gain by using the URL instead of the JNLPFile?

core/src/main/java/net/sourceforge/jnlp/Launcher.java
```diff
@@ -137,33 +110,12 @@ private Launcher(LaunchHandler handler) {
...
-        this.updatePolicy = policy;
-    }
-
-    /**
-     * @return the update policy used when launching applications.
-     */
-    public UpdatePolicy getUpdatePolicy() {
-        return updatePolicy;
+    public Launcher(final LaunchHandler handler, final UpdatePolicy policy, boolean exitOnFailure) {
+        this.handler = (handler == null) ? JNLPRuntime.getDefaultLaunchHandler() : null;
```
sclassen: I believe this should not assign `null` but instead `handler`
Also parentheses can be removed, but maybe leaving them in is better for readability, I'm undecided

core/src/main/java/net/sourceforge/jnlp/Launcher.java
```diff
@@ -376,74 +311,16 @@ private void addArguments(JNLPFile file, List<String> args) {
...
-        List<String> javawsArgs = new LinkedList<String>();
-        javawsArgs.add(url.toString());
-        launchExternal(new LinkedList<String>(), javawsArgs);
-    }
-
     /**
      * Launches the JNLP file at the specified location in a new JVM
      * instance. All streams are properly redirected.
-     * @param vmArgs the arguments to pass to the jvm
+     * @param file the JNLP file to read arguments an JVM details from
```
sclassen: typo: should be `and` JVM details

core/src/main/java/net/sourceforge/jnlp/Launcher.java
```diff
@@ -518,17 +395,17 @@ protected ApplicationInstance launchApplication(JNLPFile file) throws LaunchExce
                 return null;
             }
 
-            if (JNLPRuntime.getForksAllowed() && file.needsNewVM()) {
+            if (JNLPRuntime.getForksAllowed() /*&& file.needsNewVM()*/) {
```
sclassen: delete commented code

PR #331, opened 2019-07-25
fix systemjar detection that fails wrongly when path case is different

AndreasEhret:
depending on which OS and how Boot is called system jars are not detected as such and this do not get allPermissions but end up in a checkPermission loop.

Screenshot 2019-07-25 at 10 35 21


AndreasEhret:
we currently discussing an improved solution for this, so I will provide an updated Pull Request on this

PR #352, opened 2019-08-02
backport of #340

lherschi:


netx/net/sourceforge/jnlp/security/HttpsCertVerifier.java
```diff
@@ -184,10 +184,6 @@ private String getNamesForCert(X509Certificate c) {
                     }
                 }
             }
-
-            if (subjAltNames != null)
```
sclassen: I agree that the lines removed in this PR are incorrect.
But simply removing this neglects the case when `derValue.getAsString()` returns an empty string.

lherschi:
Ok, then I fix that first in the master branch and then make a new backport.

lherschi:
I can see right now that master has not needed a fix since the code cleanup of HttpsCertVerifier (07724b0)

PR #355, opened 2019-08-05
review of #350

hendrikebbers:
Since #350 was merged before I did a review I put some final changes here in the PR:
- final added
- null checks
- switched 1 method from private -> public since I would like to use it

### DIFF
--- a/common/src/main/java/net/adoptopenjdk/icedteaweb/jnlp/version/SimpleRange.java
+++ b/common/src/main/java/net/adoptopenjdk/icedteaweb/jnlp/version/SimpleRange.java
@@ -47,9 +47,9 @@ class SimpleRange {
     private final VersionId versionId;
     private final VersionModifier modifier;
 
-    private SimpleRange(VersionId versionId, VersionModifier modifier) {
-        this.versionId = Assert.requireNonNull(versionId, "versionId");
-        this.modifier = Assert.requireNonNull(modifier, "modifier");
+    private SimpleRange(final VersionId versionId, final VersionModifier modifier) {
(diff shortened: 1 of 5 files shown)


common/src/main/java/net/adoptopenjdk/icedteaweb/jnlp/version/SimpleRange.java
```diff
@@ -48,8 +48,8 @@
     private final VersionModifier modifier;
 
     private SimpleRange(VersionId versionId, VersionModifier modifier) {
-        this.versionId = versionId;
-        this.modifier = modifier;
+        this.versionId = Assert.requireNonNull(versionId, "versionId");
```
sclassen: private method only called from within this class and calling code never passes null
do we still need the checks?

common/src/main/java/net/adoptopenjdk/icedteaweb/jnlp/version/VersionRange.java
```diff
@@ -48,9 +48,9 @@
     private final String versionRange;
     private final SimpleRange[] ranges;
 
-    private VersionRange(String versionId, SimpleRange[] ranges) {
+    private VersionRange(final String versionId, final SimpleRange[] ranges) {
         this.versionRange = versionId;
-        this.ranges = ranges;
+        this.ranges = Assert.requireNonNull(ranges, "ranges");
```
sclassen: same as above, calling code never passes null

sclassen:
@hendrikebbers 
did some more code cleanup. feel free to merge this if you think it is OK

PR #363, opened 2019-08-13
Add the jar file name to exception message

hendrikebbers:
If you start an application. with several JARs and only 1 is corrupt you currently do not see any info about the file. This PR adds the name off the JAR to the error message

### DIFF
--- a/core/src/main/java/net/sourceforge/jnlp/tools/JarCertVerifier.java
+++ b/core/src/main/java/net/sourceforge/jnlp/tools/JarCertVerifier.java
@@ -25,7 +25,6 @@
 
 package net.sourceforge.jnlp.tools;
 
-import net.adoptopenjdk.icedteaweb.IcedTeaWebConstants;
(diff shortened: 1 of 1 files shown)


core/src/main/java/net/sourceforge/jnlp/tools/JarCertVerifier.java
```diff
@@ -302,7 +302,7 @@ private VerifyResult verifyJar(String jarName) throws Exception {
 
         } catch (Exception e) {
             LOG.error(IcedTeaWebConstants.DEFAULT_ERROR_MESSAGE, e);
```
sclassen: Could you also improve the error message in the log?

core/src/main/java/net/sourceforge/jnlp/tools/JarCertVerifier.java
```diff
@@ -302,7 +302,7 @@ private VerifyResult verifyJar(String jarName) throws Exception {
...
-            throw e;
```
sclassen: Can you include the original exception as the cause?

PR #366, opened 2019-08-14
Launchers now uses assembled javaws.jar

judovana:
- portable build is on fully assembled jar with dependencies
- distribution build is on jar with assembled ITW only, with deps expected on system

sclassen, never thaked you, nor appreciate denough the assmbled jars. They indeed made this much better. I would like to keep itw images as close to karakun release as possible. Feel free to criticise.


launchers/configure.sh
```diff
@@ -71,25 +71,12 @@ else
   readonly ICO_TARGET_DIR=$ICO_TARGET_DIR
 fi
 
-if [ "x$SNAPSHOT" == "x" ] ; then
```
sclassen: The comparison on line 74 is implemented different then on line 92.
My shell knowledge is more on a 'monkey see - monkey do' level. So I am not sure if both of them behave exactly the same.
Never the less I would suggest to stick with only one approach to keep things consistent

judovana:
Thank you. I will keep it consistent and update PR

PR #376, opened 2019-08-23
AppletDesc now accept also class.class suffixed  entry point in its applet-desc xml element

judovana:
https://github.com/AdoptOpenJDK/IcedTea-Web/issues/361

      appelts can have class specified both pkg.myclass and pkg.myclass.class
      There are several places in ITW where this can be applied
      I had decided to go eith direct getter from AppletDesc for several reasons:
      - it is applet-desc specific, and application-desc should not be crippeld by this
      - in checkForMain is palce where .class is appended, that would need to be adjsuted
      - checkEntryPoint is comparin main class against record in manifest.
      -- classes in manifest are supposed to be suffix .class free. Although you can smuggle class.class into manifest, that is usually bringing you toubles.
      -- eg. java -jar willnot load such main-class
      - I can be wrong with the manifest usage

### DIFF
--- a/core/src/main/java/net/adoptopenjdk/icedteaweb/jnlp/element/application/AppletDesc.java
+++ b/core/src/main/java/net/adoptopenjdk/icedteaweb/jnlp/element/application/AppletDesc.java
@@ -127,7 +127,28 @@ public class AppletDesc implements EntryPoint {
      */
     @Override
     public String getMainClass() {
-        return mainClass;
+        return cleanMainClass();
+    }
+
+    /**
+     * https://github.com/AdoptOpenJDK/IcedTea-Web/issues/361
+     * appelts can have class specified both pkg.myclass and pkg.myclass.class
+     * There are several places in ITW where this can be applied
+     * I had decided to go eith direct getter from AppletDesc for several reasons:
+     * - it is applet-desc specific, and application-desc should not be crippeld by this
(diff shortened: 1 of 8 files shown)


judovana:
hmm. Looks like I broke:
[ERROR] Errors: 
[ERROR]   PluginBridgeTest.teardown:101 » OverlappingFileLock
[ERROR]   PluginBridgeTest.testDeadLock:118 » OverlappingFileLock
[ERROR]   ResourceTrackerTest.teardownDownloadServer:262 » OverlappingFileLock
[ERROR]   BootTest.checkMainMethondInAppletDesc:152 » OverlappingFileLock
[INFO] 
[ERROR] Tests run: 1058, Failures: 0, Errors: 4, Skipped: 55


will check that

judovana:
Patch itself is good, but the two test I added are bad. They break the environment (however i'm unsure why so much)  I will try to transform them to your new integration world rather

judovana:
have replaced the unit test by integration tests. Well havd few issues with that engine, but is defifnitley interesting. TY!

sclassen:
yes, I don't like the part in the pom.xml which needs to be duplicated all the time.
But so far I have cannot think of another approach using maven.
I honestly start to think of using gradle or ant just to do the packaging of the jars...

judovana:
That can still leave impossible and tricky to make signed jars, or tempted jars. But well,

PR #377, opened 2019-08-23
add  pack.gz and .gz conditionally

### LINKED ISSUE #374: Fix for Packed resources are stored in cache with double file extension (#12)  had broken phetsims
commit of https://github.com/AdoptOpenJDK/IcedTea-Web/commit/3f08fb429347209a8d7d8f2dde (from #12) had broken  http://phetsims.colorado.edu/sims/circuit-construction-kit/circuit-construction-kit-dc_en.jnlp

judovana:
Fixes #374

btw, I do not know if this is correct fix! Comment from @lherschi would be awesom

### DIFF
--- a/core/src/main/java/net/sourceforge/jnlp/cache/ResourceDownloader.java
+++ b/core/src/main/java/net/sourceforge/jnlp/cache/ResourceDownloader.java
@@ -345,9 +345,17 @@ public class ResourceDownloader implements Runnable {
             // return ".gz", so if we check gzip first, we would end up
             // treating a pack200 file as a jar file.
             if (packgz) {
-                downloadPackGzFile(connection, downloadFrom, downloadTo);
+                if (downloadFrom.getFile().endsWith(".pack.gz")) {
+                    downloadPackGzFile(connection, downloadFrom, downloadTo);
+                } else {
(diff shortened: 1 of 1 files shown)


sclassen:
fixes #374

core/src/main/java/net/sourceforge/jnlp/cache/ResourceDownloader.java
```diff
@@ -345,9 +345,17 @@ private void downloadResource() {
             // return ".gz", so if we check gzip first, we would end up
             // treating a pack200 file as a jar file.
             if (packgz) {
-                downloadPackGzFile(connection, downloadFrom, downloadTo);
+                if (downloadFrom.getFile().endsWith(".pack.gz")) {
+                    downloadPackGzFile(connection, downloadFrom, downloadTo);
+                } else {
+                    downloadPackGzFile(connection, new URL(downloadFrom + ".pack.gz"), downloadTo);
```
sclassen: Shouldn't we be adding the `+ ".pack.gz"` to the `downloadTo`??

sclassen: Never mind. The code is right.
I got confused by the names.

Issue #371, opened 2019-08-21
MSI package missing "application/x-java-jnlp-file" mimetype association


Pagallac:
The Windows installer is missing MimeType associations for .jnlp, causing some downloads of .jnlp files to not have a file association, if the webpage has the download Content-Type specified as "Application/x-java-jnlp-file".
![image](https://user-images.githubusercontent.com/54354267/63420487-0dfcd780-c407-11e9-882a-d5d7c5f8be00.png)

Oracle Java creates this association in the windows registry under:
HKEY_CLASSES_ROOT\MIME\Database\Content Type\application/x-java-jnlp-file

Manually creating this MimeType association after IcedTea webstart (and adoptopenjdk) installation, fixes the problem.

judovana:
that msi is poosly maintained. Please submit a patch if you can.,

middletonk:
Apologies for the necro post but I can repro this issue in the AdoptOpenJDK 8u292 MSI when using INSTALLLEVEL=3 (includes IcedTea-Web) and also with explicit naming of features. What's the best protocol for this? Reopen existing issue or file a new if proven to be generally reproducible?

douph1:
IcedTea installer is not the same as AdoptOpenJDK installer. Please fill a new bug in https://github.com/adoptium/installer/issues

douph1:
For this installer it was addressed in https://github.com/AdoptOpenJDK/IcedTea-Web/pull/153 in 1.8.1 in April 2019
This installer is available at : https://adoptopenjdk.net/icedtea-web.html

middletonk:
Shall be done. Appreciate it!

PR #384, opened 2019-08-27
msi-installer: add mimetype association, see #371

lherschi:


judovana:
Thank you for taking care for windows. Looks good.

PR #391, opened 2019-09-03
remove PluginBrige and all code which becomes unreachable by doing so

### LINKED ISSUE #333: with applet gone remove --html
This swithc have become useless, even worse, will end in error. The switch and its impl should be gone

sclassen:
fixes #333


judovana:
hi! applet-desc still keeps working right?
One  nit - applets, unlike regular javaws were able to load .class direclty.  I'm not sure uf applet-desc should support it. if not, this possible hole hsould be gone from jnlpclasslaoder to.

judovana:
There were added several integration tests with appelt-desc recently. Can you rebase and trry?

sclassen:
Merged master back into this branch.
But do not rush merging this, I will give it also another thought

sclassen:
@judovana 
I looked at it again and it seems reasonable for me. Especially since the entire applet part has been removed from Java11...
So everybody who needs applet must stay on a java version before 11 and can therefore use the Oracle Web Start if they want to

judovana:
yup. ok to go the, TY!

PR #396, opened 2019-09-04
Control Center branding

hendrikebbers:
For OpenWebStart we would love to change the header of the control Center (without removing IcedTeaWeb but changing it to something like "OpenWebStart - powered by IcedTeaWeb"). based on this I created an interface that provides all the properties and panels that we might change for OpenWebStart. By doing so we can easily provide a custom implementation for OpenWebStart.


judovana:
I like this change! Will the OpenWebStart  implementation of this interface be pushed into ITW, and so enable-able during build,   or will yor deployment process,   simply during build,  replace the in-tree provider+resources  by your implementation?

judovana:
Will you continue by your custom splash?-)

hendrikebbers:
Currently I assume that we implant the interface in the OWS Repo and just call a copied main method

PR #403, opened 2019-09-05
center control center on screen

hendrikebbers:
- center control center on screen
- allow to create instance in OWS


judovana:
there is utiliti method screenfinder. Which do exaclty what you do. You should reuse the code in that class.

judovana:
> there is utiliti method screenfinder. Which do exaclty what you do. You should reuse the code in that class.

there is utiliti class of screenfinder. Which methdos do  exactly what you do. You should reuse the code in that class.

PR #404, opened 2019-09-05
Fixed forking issues

Pawlost:
Commits to blame:
https://github.com/AdoptOpenJDK/IcedTea-Web/commit/30f3b04556be5c9289663c8ac6d57539b1fbbc85

https://github.com/AdoptOpenJDK/IcedTea-Web/commit/461969132400174241f825034b6f78065e44ebb6

### DIFF
--- a/core/src/main/java/net/sourceforge/jnlp/Launcher.java
+++ b/core/src/main/java/net/sourceforge/jnlp/Launcher.java
@@ -393,7 +393,7 @@ public class Launcher {
                 return null;
             }
 
-            if (JNLPRuntime.getForksAllowed()) {
+            if (JNLPRuntime.getForksAllowed() && file.needsNewVM()) {
                 if (!JNLPRuntime.isHeadless()){
                     SplashScreen sp = SplashScreen.getSplashScreen();
                     if (sp!=null) {


sclassen:
can you describe how this bug manifests?

judovana:
Obviously, the fork was broken.   Since the commits in blame, when it was commented and later removed, the  forking was made alway, and not only when needed.  Now it is correct again, and JVm forks only if jvmargas  (and no Xnofork)  are present.

judovana:
I'm ok with this changed, actually it is fixing severe bug.  But @Pawlost  please fix the commit message.  and add some description. The boot is good. Forking is what isa broken.

sclassen:
we depend on forking always (unless -Xnofork)
Otherwise we cannot change the Java Runtime

judovana:
But then you are doing it wrong.   Anotther way how tro fork whne you need have to be  fond. Ex -XenforceFork ? Or fork when -XanotherVm==/pathto/vm ?

sclassen:
we only decide on runtime if we need to fork, cannot pass this as a argument.
- we check the java version specified in the JNLP
- we check the JVMs installed on the current machine
- determine the best match and then fork to selected JVM

sclassen:
how can always forking introduce a bug?
the only side effect is that we start a second JVM with -Xnofork which will then not fork another JVM

judovana:
> we only decide on runtime if we need to fork, cannot pass this as a argument.
>     * we check the java version specified in the JNLP
>     * we check the JVMs installed on the current machine
>     * determine the best match and then fork to selected JVM

But this is obvious additin to method of needsNewVM isnt it? So not alwys for, but when  above conditions are met.

The unnecessary forking of JVM is evil. JVM is not exactly friendly process, and except  unexpect big process, there are the issues like duplicated error dialogue or so. 

I Agree with your conidtions to fork, but keep disagreein for 100% fork.

Pawlost:
I found another issue related to this commit: https://github.com/AdoptOpenJDK/IcedTea-Web/commit/461969132400174241f825034b6f78065e44ebb6
It is the second edit. When I am launching Boot class in IntelliJ IDE with link to the html jnlp source and Xnofork argument it causes StackOverFlow exeption. When I remove negation to condition above and skip security and policy initialization, it works.

sclassen:
yes this is known. The reason is that Intellij packs a ton of other jars onto the classpath which will result in multiple classloaders which then create a stack overflow as JNLPClassloader calls getParent() while determine if the current thread is executing ITW or a JNLP appliaction code.
But getParent() triggers a new checkPermission which triggers a getParent() and you end in an endless loop -> stackOverflow

judovana:
Hi  @Pawlost I think the fix  with not you are mentioning, is simply disbaling the security check.

@sclassen  thank you for very educative explanation!

Pawlost:
ok thanks for explanation

PR #405, opened 2019-09-05
decative specific panels in a style

hendrikebbers:
after this PR OWS can easily deactivate ITW panels in the Control Center and add new panels.

### DIFF
--- a/core/src/main/java/net/adoptopenjdk/icedteaweb/client/controlpanel/ControlPanel.java
+++ b/core/src/main/java/net/adoptopenjdk/icedteaweb/client/controlpanel/ControlPanel.java
@@ -86,7 +86,7 @@ public class ControlPanel extends JFrame {
         this.config = config;
 
         JPanel topPanel = style.createHeader();
-        JPanel mainPanel = createMainSettingsPanel();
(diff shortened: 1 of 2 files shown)


core/src/main/java/net/adoptopenjdk/icedteaweb/client/controlpanel/ControlPanel.java
```diff
@@ -163,12 +163,13 @@ private JPanel createButtonPanel() {
...
      */
-    private JPanel createMainSettingsPanel() {
+    private JPanel createMainSettingsPanel(final ControlPanelStyle style) {
+        Assert.requireNonNull(style, "style");
         final Map<String, ControlPanelProvider> providers = new HashMap<>();
         final ServiceLoader<ControlPanelProvider> serviceLoader = ServiceLoader.load(ControlPanelProvider.class);
         serviceLoader.iterator().forEachRemaining(p -> {
             final String name = p.getName();
-            if (p.isActive()) {
+            if (p.isActive() && style.isPanelActive(p.getName())) {
```
sclassen: is the p.isActive() still needed?
The extension can also not register the service to the service loader which would have the same effect...

PR #406, opened 2019-09-05
no source-jar or javadoc-jar

hendrikebbers:
Disable generation of JavaDoc by default.
A specialized profile `javaDoc` has been added to allow generation of javaDoc if required


judovana:
I agree with the reasoning. Maybe move them to separate profiles? Anyway I'm ok with compelte removal.

sclassen:
At least sources is needed if you want to be able to debug ITW fat jar. And it only takes very short time to generate sources jar

sclassen:
I am ok with killing javadoc from default but would allow to activate it with a maven profile

judovana:
Good point with attached debugger. That was reason why simialr functionality was also in original netx.

hendrikebbers:
Ok with profile. Will start vacation in 5 min so can not continue here the next 2 weeks..,

Von meinem iPhone gesendet

PR #408, opened 2019-09-05
add forkingStrategy

sclassen:
converting the forksAllow to an enum allows for taking care of all possible states.
Also for the ones of OpenWebStart


AndreasEhret:
nice solution, LGTM

judovana:
Why the conditions in https://github.com/AdoptOpenJDK/IcedTea-Web/pull/404#issuecomment-528322754  are not evaluated?  Anyway thank you for fixing the forking.

sclassen:
this will happen in openwebstart code and we will set the correct strategy on JNLPRuntime

judovana:
hmhhm. ok then.

PR #411, opened 2019-09-05
read from stdout/stderr in separate thread to prevent hanging of process

sclassen:
On windows a process will hang if either stdout or stderr buffer is full.
The process will wait indefinitely until some other process will read from the buffer to make space for more content

This PR adds code which will actively read stdout and stderr from processes

### DIFF
--- a/common/src/main/java/net/adoptopenjdk/icedteaweb/ProcessUtils.java
+++ b/common/src/main/java/net/adoptopenjdk/icedteaweb/ProcessUtils.java
@@ -37,10 +37,15 @@ exception statement from your version.
 
 package net.adoptopenjdk.icedteaweb;
 
+import net.adoptopenjdk.icedteaweb.io.IOUtils;
 import net.adoptopenjdk.icedteaweb.logging.Logger;
 import net.adoptopenjdk.icedteaweb.logging.LoggerFactory;
 
+import java.io.IOException;
+import java.io.InputStream;
 import java.util.Objects;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.Future;
 
 public class ProcessUtils {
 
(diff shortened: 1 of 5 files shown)


judovana:
Are headless dialogues still working?

sclassen:
How can I test headless dialogues?
I was thinking that in headless mode no dialogues are shown ???

judovana:
In headless mode, dialogs are printed to stdout, and are waiting for YES/NO or RYES/RNO (R like remember)  from stdin.

common/src/main/java/net/adoptopenjdk/icedteaweb/ProcessUtils.java
```diff
@@ -69,4 +74,70 @@ public static void waitForSafely(final Process process) {
...
+     * @param process the process to read the stderr from
+     * @return a future holding the content of the stderr
+     */
+    public static Future<String> readStdErrAsUtf8(final Process process) {
+        return readStreamAsUtf8(process.getErrorStream());
+    }
+
+    private static Future<String> readStreamAsUtf8(final InputStream in) {
+        final CompletableFuture<String> result = new CompletableFuture<>();
+        new Thread(() -> {
```
hendrikebbers: Please use executors (like Executors.newChachedThreadPool) instead of creating a new thread al the time

common/src/main/java/net/adoptopenjdk/icedteaweb/ProcessUtils.java
```diff
@@ -69,4 +74,70 @@ public static void waitForSafely(final Process process) {
...
+            try {
+                final String content = IOUtils.readContentAsUtf8String(in);
+                result.complete(content);
+            } catch (IOException e) {
+                LOG.error(IcedTeaWebConstants.DEFAULT_ERROR_MESSAGE, e);
```
hendrikebbers: Please add more specific message

common/src/main/java/net/adoptopenjdk/icedteaweb/ProcessUtils.java
```diff
@@ -69,4 +74,70 @@ public static void waitForSafely(final Process process) {
...
+                throw new RuntimeException(e);
```
hendrikebbers: And add the message here ;)

judovana:
IS this stuck somewhere?

PR #416, opened 2019-09-11
simplify code to handle splashscreen via environment variables

AndreasEhret:


core/src/main/java/net/adoptopenjdk/icedteaweb/client/parts/splashscreen/SplashUtils.java
```diff
@@ -169,52 +183,42 @@ static SplashErrorPanel getErrorSplashScreen(int width, int height, SplashUtils.
...
      * @param splashReason
      * @param loadingException
      * @param isError
      */
-    public static SplashPanel getSplashScreen(int width, int height, SplashUtils.SplashReason splashReason, Throwable loadingException, boolean isError) {
+    public static SplashPanel getSplashScreen(final int width, final int height, final SplashUtils.SplashReason splashReason, final Throwable loadingException, final boolean isError) {
+        SplashPanel splashPanel;
         String splashEnvironmentVar = null;
-        String pluginSplashEnvironmentVar = null;
+
```
sclassen: If you extract `String splashEnvironmentVar = null` togehter with the try block into a new method you can make the splashEnvironmentVar final

AndreasEhret: extracted to method

PR #422, opened 2019-09-13
Cleanup of various cache related artifacts

AndreasEhret:
- refactored LRU cache
- extract Windows shortcut handling
- removed duplicated stream copy from CacheUtil


core/src/main/java/net/sourceforge/jnlp/cache/CacheLRUWrapper.java
```diff
@@ -287,24 +287,15 @@ public synchronized boolean updateEntry(String oldKey) {
...
-            @Override
-            public int compare(Entry<String, String> e1, Entry<String, String> e2) {
-                Long t1 = Long.parseLong(e1.getKey().split(",")[0]);
-                Long t2 = Long.parseLong(e2.getKey().split(",")[0]);
-
-                int c = t1.compareTo(t2);
-                return c < 0 ? 1 : (c > 0 ? -1 : 0);
-            }
-        });
+        entries.sort(CacheLRUWrapper::compare);
```
sclassen: `entries.sort(Comparator.comparingLong(e -> Long.parseLong(e.getKey().split(",")[0])));`

core/src/main/java/net/sourceforge/jnlp/cache/CacheLRUWrapper.java
```diff
@@ -347,7 +328,7 @@ public synchronized boolean containsValue(String value) {
      * @param path Path to generate a key with.
      * @return String representing the a key.
      */
-    public String generateKey(String path) {
+    String generateKey(String path) {
```
sclassen: Maybe the key could be a class of its own and not just a string...

core/src/main/java/net/sourceforge/jnlp/cache/CacheLRUWrapper.java
AndreasEhret: nice one. Done

PR #444, opened 2019-10-02
New ResourceBundle can be defined

hendrikebbers:
Based on this change we can use the translation in the OpenWebStart dialogs without adding unused translation resources in ITW.


sclassen:
Somehow your change breaks the tests.
I did an alternative implementation which encapsulates the chaining of resource bundles within the translator
see #450

hendrikebbers:
@sclassen have not seen your comment. Will have a look

hendrikebbers:
Closed since we recreated the feature in https://github.com/AdoptOpenJDK/IcedTea-Web/pull/450

PR #449, opened 2019-10-02
Batch launcher catchup

AurumTheEnd:


judovana:
Nothing bad arround. I'm unable to judge bats, Lauren is. I will leave this hanging for a while, maybe some win guru will check.

thank you very much for fixing this!

judovana:
Changed to bug, the bats, as copied from 1.8 were not working since mavenization.

PR #450, opened 2019-10-03
add chained resource bundles and allow to add bundles to the translator

sclassen:


hendrikebbers:
I do not see the benefit against https://github.com/AdoptOpenJDK/IcedTea-Web/pull/444 and https://github.com/karakun/OpenWebStart/pull/43

RessourceBundle can hold a parent and based not this the following simple workflow can be done: ITW -> only one bundle without parent
OWS -> OWS bundle with ITW bundle as parent

Next to this the PR do not change the modifier of the translate methods to public. By doing so you can not do:

```
Translator t = Translator.getInstance();
t.translate("a.b.1);
t.translate("a.b.2);
t.translate("a.b.3);
t.translate("a.b.4);
```

PR #454, opened 2019-10-08
remove ClassFinder.findAllMatchingTypes()

sclassen:
The class finder is used to load ALL classes in order to find the ones which implement a specific interface.
It may not be as future proof as the reflective search by the ClassFinder but this approach is extremely simple and does not need to load all classes in the ClassPath


core/src/main/java/net/adoptopenjdk/icedteaweb/client/parts/dialogs/security/remember/RememberableDialog.java
```diff
@@ -38,15 +38,37 @@
...
-    public DialogResult getValue();
-    public JNLPFile getFile();
-    public DialogResult readValue(String s);
-    
+    RememberPanelResult getRememberAction();
+    DialogResult getValue();
+    JNLPFile getFile();
+    DialogResult readValue(String s);
+
+    List<Class<? extends RememberableDialog>> allRememberableDialog = Collections.unmodifiableList(Stream.of(
```
hendrikebbers: Since this is a constant the name should be upper case

PR #456, opened 2019-10-08
add @NotThreadSafe as hints to surfier test runner

sclassen:
separating the Java changes from the Travis changes in #451 
Thanks @Artoria2e5 for the valuable contribution


sclassen:
@hendrikebbers 
I added the dependency to the test-extension. That is why it must not be in scope `test` as the test-extension itself is a `test` scoped dependency

PR #466, opened 2019-10-15
Stuck downloadingresources

knickedixen:
We've had some problems with downloads getting stuck when starting an application. Problem seemed to be that IcedTea-Web started too many connections in a short amount of time, up to 100 in our case. This seemed to cause the network or firewall to freak out a bit.

Limiting the amount of parallel downloads solved the problem for us. Three is an arbitrary number we just made up, but it seemed to solve the problem without increasing the total load time. We had to split initializing and running the download to fix the progress indicator from going crazy.

### DIFF
--- a/core/src/main/java/net/sourceforge/jnlp/cache/ResourceDownloader.java
+++ b/core/src/main/java/net/sourceforge/jnlp/cache/ResourceDownloader.java
@@ -77,17 +77,46 @@ class ResourceDownloader implements Runnable {
         }
 
         if (!isProcessing) {
-            CachedDaemonThreadPoolProvider.DAEMON_THREAD_POOL.execute(new ResourceDownloader(resource, lock));
+            final ResourceDownloader rd = new ResourceDownloader(resource, lock);
+            CachedDaemonThreadPoolProvider.DAEMON_THREAD_POOL.execute(
+                new Runnable() {
+                    @Override
+                    public void run() {
+                        rd.runInitialize();
+                        CachedDaemonThreadPoolProvider.DAEMON_THREAD_POOL.execute(
+                            new Runnable() {
+                                @Override
+                                public void run() {
+                                    rd.runDownload();
+                                }
+                            });
+                    }
+                });
         }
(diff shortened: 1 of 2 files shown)


core/src/main/java/net/sourceforge/jnlp/cache/ResourceDownloader.java
```diff
@@ -77,17 +77,46 @@ static void startDownload(Resource resource, final Object lock) {
         }
 
         if (!isProcessing) {
-            CachedDaemonThreadPoolProvider.DAEMON_THREAD_POOL.execute(new ResourceDownloader(resource, lock));
+            final ResourceDownloader rd = new ResourceDownloader(resource, lock);
+            CachedDaemonThreadPoolProvider.DAEMON_THREAD_POOL.execute(
+                new Runnable() {
```
hendrikebbers: Please use Lambda expression

core/src/main/java/net/sourceforge/jnlp/cache/ResourceDownloader.java
```diff
@@ -77,17 +77,46 @@ static void startDownload(Resource resource, final Object lock) {
...
+                    @Override
+                    public void run() {
+                        rd.runInitialize();
+                        CachedDaemonThreadPoolProvider.DAEMON_THREAD_POOL.execute(
```
hendrikebbers: Why is the download done in a separate runnable?

sclassen: I like that the initialization is done in a separate thread from the actual download.
But the code needs to be cleaned up.
@hendrikebbers if it is OK for you we can accept the PR and do the code cleanup afterwards

hendrikebbers: Fine with that. Will check if I have some time later today or tomorrow. In that case I will merge the PR and directly do a new one for the CleanUp.

PR #478, opened 2019-10-23
Functionallity to disable Ui elements

hendrikebbers:
Elements (like textfields or Checkboxes) in the UI can be disabled by a configuration property.

### DIFF
--- a/core/src/main/java/net/adoptopenjdk/icedteaweb/client/util/UiLock.java
+++ b/core/src/main/java/net/adoptopenjdk/icedteaweb/client/util/UiLock.java
@@ -1,0 +1,43 @@
+package net.adoptopenjdk.icedteaweb.client.util;
+
+import net.adoptopenjdk.icedteaweb.Assert;
+import net.sourceforge.jnlp.config.DeploymentConfiguration;
+
+import javax.swing.JComboBox;
+import javax.swing.text.JTextComponent;
+import java.awt.Component;
+import java.util.Optional;
+
+/**
+ * Helper class that can be used to check if a UI element should be locked / disabled in the UI
+ */
+public class UiLock {
+
+    private final DeploymentConfiguration deploymentConfiguration;
+
+    private final static String LOCK_SUFFIX = ".locked";
+
+    public UiLock(final DeploymentConfiguration deploymentConfiguration) {
+        this.deploymentConfiguration = Assert.requireNonNull(deploymentConfiguration, "deploymentConfiguration");
+    }
+
+    public void update(final String propertyName, final Component component) {
+        final boolean isLocked = Optional.ofNullable(deploymentConfiguration.getProperty(propertyName + LOCK_SUFFIX))
+                .map(s -> Boolean.parseBoolean(s))
+                .orElse(false);
+        if(isLocked) {
+            component.setEnabled(false);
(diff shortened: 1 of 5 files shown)


core/src/main/java/net/adoptopenjdk/icedteaweb/client/util/UiLock.java
```diff
@@ -0,0 +1,43 @@
...
+    private final DeploymentConfiguration deploymentConfiguration;
+
+    private final static String LOCK_SUFFIX = ".locked";
+
+    public UiLock(final DeploymentConfiguration deploymentConfiguration) {
+        this.deploymentConfiguration = Assert.requireNonNull(deploymentConfiguration, "deploymentConfiguration");
+    }
+
+    public void update(final String propertyName, final Component component) {
+        final boolean isLocked = Optional.ofNullable(deploymentConfiguration.getProperty(propertyName + LOCK_SUFFIX))
```
sclassen: Would be better to add a method `isLocked(String)` to the DeploymentConfiguration and make use of the `currentConfiguration.get(key).isLocked()` which is available inside of DeploymentConfiguration

core/src/main/java/net/adoptopenjdk/icedteaweb/client/util/UiLock.java
```diff
@@ -0,0 +1,43 @@
...
+            if(component instanceof JTextComponent) {
+                ((JTextComponent) component).setEditable(false);
+            }
+            if(component instanceof JComboBox) {
+                ((JComboBox) component).setEditable(false);
+            }
+        }
+    }
+
+    public void addLock(final String propertyName) {
```
sclassen: never used

sclassen: OK, I found the usage in OWS.
Can you please do the same as with `isLocked` and add a helper to the DeploymentConfiguration so we can keep the knowledge about how a property is locked within the DeploymentConfiguration.

sclassen: And then finally inline this method as it is only delegating the call to DeploymentConfiguration...

core/src/main/java/net/adoptopenjdk/icedteaweb/client/util/UiLock.java
hendrikebbers: What do you mean? `DeploymentConfiguration` has a `getProperty(key)` method that returns a string. There is not `get(key)` method.

core/src/main/java/net/adoptopenjdk/icedteaweb/client/util/UiLock.java
hendrikebbers: Since the class is already used in ITW I do not see how we not "keep the knowledge". I could add JavaDoc to the new class ;)
Adding everything to `DeploymentConfiguration`: I do not think that this is a good solution since the lock functionality is only used in the UI and the `DeploymentConfiguration` is used all over the place. Therefore I like the idea to have a facade / wrapper for the lock functionality instead of adding more functionality to the `DeploymentConfiguration` class

PR #481, opened 2019-10-24
Maven/eclipse improvements

don-vip:
Some fixes/tweaks to get a clean git repository and less warnings in Maven build.


pom.xml
```diff
@@ -119,6 +121,7 @@
                         <environmentVariables>
                             <ICEDTEA_WEB_SPLASH>DEFAULT</ICEDTEA_WEB_SPLASH>
                         </environmentVariables>
+                        <redirectTestOutputToFile>true</redirectTestOutputToFile>
```
sclassen: The output is now much cleaner.
Problem is, in travis we can no longer see why a test may have failed because all output is gone

sclassen:
I Like the changes, thanks
@hendrikebbers what do you think about the `<redirectTestOutputToFile>` ??

pom.xml
don-vip: Does this project own an S3 bucket? This way we could upload the text files as artifacts

hendrikebbers: At the moment we have no good way to store the output. Please remove that line so that the PR can be merged :)

don-vip: Done! :)

PR #485, opened 2019-10-29
first implementation of PrioritizedParallelExecutor

sclassen:


core/src/main/java/net/adoptopenjdk/icedteaweb/resources/downloader/PrioritizedParallelExecutor.java
```diff
@@ -0,0 +1,218 @@
...
+     * The callables must be sorted by priority with the highest priority as the first in list.
+     * Executions of lower priority are canceled as soon as a higher priority result has successfully been calculated.
+     * The returned {@link Future} will complete as soon as the first callable completed successfully and all other
+     * callables with higher priority (if any) have completed exceptionally.
+     *
+     * @param callables the callables to execute.
+     * @return a future holding the best result from all callables.
+     */
+    public <V> Future<V> getSuccessfulResultWithHighestPriority(List<Callable<V>> callables) {
+        if (callables == null || callables.isEmpty()) {
```
hendrikebbers: Maybe you do an assert first and than do the empty check. Otherwise you do not know if it is null or empty based on the exception

core/src/main/java/net/adoptopenjdk/icedteaweb/resources/downloader/PrioritizedParallelExecutor.java
```diff
@@ -0,0 +1,218 @@
...
+
+        @Override
+        public V call() throws Exception {
+            final V result = delegate.call();
+            cancelLowerPriority();
+            return result;
+        }
+
+        private void cancelLowerPriority() {
+            complete.set(true);
```
hendrikebbers: I would assume to set this in the call method

core/src/main/java/net/adoptopenjdk/icedteaweb/resources/downloader/PrioritizedParallelExecutor.java
```diff
@@ -0,0 +1,218 @@
...
+         * This is not passed in the constructor as the value is not known then.
+         *
+         * If this callable has already been completed successfully then the passed future will be canceled immediately.
+         *
+         * @param lowerPriority the lower priority future to be canceled after completing this callable successfully.
+         */
+        void setLowerPriority(Future<V> lowerPriority) {
+            this.lowerPriority.set(lowerPriority);
+            if (complete.get()) {
+                lowerPriority.cancel(true);
```
hendrikebbers: call cancelLowerPriority() here

core/src/main/java/net/adoptopenjdk/icedteaweb/resources/downloader/PrioritizedParallelExecutor.java
```diff
@@ -0,0 +1,218 @@
...
+         * @param lowerPriority the lower priority future to be canceled this future is canceled.
+         */
+        void setLowerPriority(Future<V> lowerPriority) {
+            this.lowerPriority.set(lowerPriority);
+
+            if (terminatedExceptionally.get()) {
+                // TODO: should the body of this if be executed in a separate thread to not block the current thread
+                //  which is about to plug together the future chain...
+                try {
+                    lowerPriorityResult.complete(lowerPriority.get());
```
hendrikebbers: by doing this the method will block until the `lowerPrioroty` is done. I would not assume something like this in a setter method

core/src/main/java/net/adoptopenjdk/icedteaweb/resources/downloader/PrioritizedParallelExecutor.java
sclassen: Can do but there is basically no difference in action to take.
It means one of the caller passed in non-sense and we need to fix it

core/src/main/java/net/adoptopenjdk/icedteaweb/resources/downloader/PrioritizedParallelExecutor.java
```diff
@@ -0,0 +1,218 @@
...
+
+        /**
+         * Set a lower priority to be canceled.
+         * This is not passed in the constructor as the value is not known then.
+         *
+         * If this callable has already been completed successfully then the passed future will be canceled immediately.
+         *
+         * @param lowerPriority the lower priority future to be canceled after completing this callable successfully.
+         */
+        void setLowerPriority(Future<V> lowerPriority) {
```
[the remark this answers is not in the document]
sclassen: private internal class only used in this file.
null is never passed

core/src/main/java/net/adoptopenjdk/icedteaweb/resources/downloader/PrioritizedParallelExecutor.java
hendrikebbers: the difference is that the exception will give you an hint if its empty or null. But this is only a minor comment and not mandatory ;)

core/src/main/java/net/adoptopenjdk/icedteaweb/resources/downloader/PrioritizedParallelExecutor.java
```diff
@@ -0,0 +1,220 @@
...
+         * Set a lower priority to be canceled.
+         * This is not passed in the constructor as the value is not known then.
+         *
+         * @param lowerPriority the lower priority future to be canceled this future is canceled.
+         */
+        void setLowerPriority(Future<V> lowerPriority) {
+            this.lowerPriority.set(lowerPriority);
+
+            if (terminatedExceptionally.get()) {
+                Executors.newSingleThreadExecutor().execute(() -> {
```
hendrikebbers: mmmh, I do not know if `Executors.newSingleThreadExecutor()` is a good idea. The basic idea behind this PR was to have a list that is executed by a concrete executor.

core/src/main/java/net/adoptopenjdk/icedteaweb/resources/downloader/PrioritizedParallelExecutor.java
```diff
@@ -0,0 +1,220 @@
...
+
+        @Override
+        public V call() throws Exception {
+            final V result = delegate.call();
+            complete.set(true);
+            cancelLowerPriority();
+            return result;
+        }
+
+        private void cancelLowerPriority() {
```
hendrikebbers: Since this is called in the `call()` method and in the `setLowerPriority(...)` method it could be called twice (even parallel) based on a race condition.  Based on this the lowerPriority can be canceled more than once. I do not know if this might end in problems.

core/src/main/java/net/adoptopenjdk/icedteaweb/resources/downloader/PrioritizedParallelExecutor.java
```diff
@@ -0,0 +1,220 @@
...
+            final Future<V> lowerPriority = this.lowerPriority.get();
```
hendrikebbers: please rename the local lowerPriority to lowerPriorityFuture. The method is very hard to read based on the same variable names

core/src/main/java/net/adoptopenjdk/icedteaweb/resources/downloader/PrioritizedParallelExecutor.java
```diff
@@ -0,0 +1,220 @@
...
+         * This is not passed in the constructor as the value is not known then.
+         * <p>
+         * If this callable has already been completed successfully then the passed future will be canceled immediately.
+         *
+         * @param lowerPriority the lower priority future to be canceled after completing this callable successfully.
+         */
+        void setLowerPriority(Future<V> lowerPriority) {
+            this.lowerPriority.set(lowerPriority);
+            if (complete.get()) {
+                cancelLowerPriority();
```
hendrikebbers: Since cancelLowerPriority calls cancel() on the future object the method might block (cancel() blocks until the future is really canceled)

core/src/main/java/net/adoptopenjdk/icedteaweb/resources/downloader/PrioritizedParallelExecutor.java
```diff
@@ -0,0 +1,220 @@
...
+        public V get(long timeout, TimeUnit unit) throws InterruptedException, ExecutionException, TimeoutException {
+            final long start = System.nanoTime();
+            try {
+                return delegate.get(timeout, unit);
+            } catch (ExecutionException e) {
+                final long elapsed = System.nanoTime() - start;
+                final long remaining = unit.toNanos(timeout) - elapsed;
+                return returnLowerPriority(remaining, NANOSECONDS);
+            } catch (TimeoutException e) {
+                return returnLowerPriority(1, MILLISECONDS);
```
hendrikebbers: Does this really makes sense to wait longer? Maybe better:

if(lowerPrio.isDone()) {return lowerPrio.get()} else {throw e;}

Next to this calling get() and calling get(long timeout, TimeUnit unit) could now return different results. I think that is a bug and should be changed.

core/src/main/java/net/adoptopenjdk/icedteaweb/resources/downloader/PrioritizedParallelExecutor.java
```diff
@@ -0,0 +1,220 @@
...
+        private void cancelLowerPriority() {
+            terminatedExceptionally.set(true);
+            final Future<V> lowerPriority = this.lowerPriority.get();
+            if (lowerPriority != null) {
+                lowerPriority.cancel(true);
+            }
+        }
+
+        private V returnLowerPriority() throws InterruptedException, ExecutionException {
+            terminatedExceptionally.set(true);
```
hendrikebbers: The name "terminatedExceptionally" does not really fit to the usage since you mutate the field only in methods that are related to the lowerPrioFuture

core/src/main/java/net/adoptopenjdk/icedteaweb/resources/downloader/PrioritizedParallelExecutor.java
```diff
@@ -0,0 +1,220 @@
...
+        }
+
+        if (callables.size() == 1) {
+            return executor.submit(callables.get(0));
+        }
+
+        final CancelingCallable<V> head = new CancelingCallable<>(callables.get(0));
+        final List<Callable<V>> tail = callables.subList(1, callables.size());
+
+        final ChainableFuture<V> result = new ChainableFuture<>(executor.submit(head));
```
hendrikebbers: This looks like you pass each Callable 2 times to the executor. First you add it in line 54 and now you add it a second time. So for 4 callable in the list we need 8 tasks runs in the executor.

sclassen: no, I only see this one submit, there is not other one

core/src/main/java/net/adoptopenjdk/icedteaweb/resources/downloader/PrioritizedParallelExecutor.java
sclassen: yes, so should I execute it in a different thread then??

core/src/main/java/net/adoptopenjdk/icedteaweb/resources/downloader/PrioritizedParallelExecutor.java
sclassen: cancel will simply return false if the future is already canceled

core/src/main/java/net/adoptopenjdk/icedteaweb/resources/downloader/PrioritizedParallelExecutor.java
sclassen: how else would you prevent the call to `setLowerPriority` from being blocking

core/src/main/java/net/adoptopenjdk/icedteaweb/resources/downloader/PrioritizedParallelExecutor.java
hendrikebbers: line 54 and line 60

core/src/main/java/net/adoptopenjdk/icedteaweb/resources/downloader/PrioritizedParallelExecutor.java
hendrikebbers: I do not say that I have the perfect solution ;) I assume that at least you should pass the executor up to this place. Do not know how much complexity this will add.

core/src/main/java/net/adoptopenjdk/icedteaweb/resources/downloader/PrioritizedParallelExecutor.java
hendrikebbers: the best case would be to not start it if this one is already completed. And this check must be locked.

But this is again the problem. As far as I can say this complete PR can end in several problems based on race conditions that we just do not see at the moment.

core/src/main/java/net/adoptopenjdk/icedteaweb/resources/downloader/PrioritizedParallelExecutor.java
sclassen: moved setters to make name more fitting

core/src/main/java/net/adoptopenjdk/icedteaweb/resources/downloader/PrioritizedParallelExecutor.java
sclassen: but they are exclusive.
Line 54:  `return executor.submit(callables.get(0));`

So if the callable is submitted there it will never be submitted on line 60 and vice-versa

hendrikebbers: ok, missed the `return`. My fault ;)

core/src/main/java/net/adoptopenjdk/icedteaweb/resources/downloader/PrioritizedParallelExecutor.java
hendrikebbers: Why is this comment resolved? `get()` and `get(long timeout, TimeUnit unit)` could still return different values.

sclassen: I did implement your change request. it checks for lowerPrio done and if so returns that value otherwise throws original exception

sclassen:
Will close this PR for now.
As it is "only" a performance improvement we can keep going without this.
Also as the branch only adds new files and does not modify any existing code there will not be any conflicts.
We can come back to this branch once there is more time.

Issue #313, opened 2019-07-05
Building launchers on Windows


daddyaddy:
We can not build the build.sh file by Linux bash (installed on Windows 10) because we have a Java version for Windows.

How to build launchers for Windows?

Sam1970:
Hi,

you can using Alex Kashchenko Dev-Kit and uninstall java on Windows 10:
https://github.com/akashche/itw_win_devkit

regards

lherschi:
This Dev-Kit uses the old mercurial repository and is therefore obsolete. You net to set the environment variable JRE to your Java runtime (your Windows drives should be automatically be mounted under /mnt) an then run mvn -Plaunchers install. But in the current master the build does not work. See https://github.com/AdoptOpenJDK/IcedTea-Web/pull/306

hendrikebbers:
@daddyaddy you can download a preview here www.openwebstart.com

judovana:
The build environment for launchers - both rust and shell - are build in cygwin. I was never trying other apporaches.

judovana:
@daddyaddy  @lherschi  is the current build working for you?  I had revisited it, and fixed few bugs. However, it is still depndent on cygwin. I belive, it will require only small chages to make it alive under linux subsystem for windows. Morover only get rid of cygoath, which is use donly on two places. Thoughts?

lherschi:
Yes, the current build works for me.

PR #490, opened 2019-10-31
#313 - Fix Windows launchers build

don-vip:
The current command in configure.sh selects two jar files (the normal one and the shaded one) and provokes a strange error when trying to build launchers on Windows.

I assume we want the shaded jar and not the normal one.

### DIFF
--- a/launchers/configure.sh
+++ b/launchers/configure.sh
@@ -75,7 +75,7 @@ if [ "x$ITW_LIBS" == "xDISTRIBUTION" ] ; then
   readonly JAVAWS_SRC=`ls $PROJECT_TOP/artifact-no-dependencies/target/icedtea-web-no-dependencies-*.jar | grep -v sources | grep -v javadoc`
   readonly JAVAWS_SRC_SRC=`ls $PROJECT_TOP/artifact-no-dependencies/target/icedtea-web-no-dependencies-*.jar | grep sources`
 else
-  readonly JAVAWS_SRC=`ls $PROJECT_TOP/artifact-all-dependencies/target/icedtea-web-all-dependencies-*.jar | grep -v sources | grep -v javadoc`
+  readonly JAVAWS_SRC=`ls $PROJECT_TOP/artifact-all-dependencies/target/icedtea-web-all-dependencies-*.jar | grep -v sources | grep -v javadoc | grep -v shaded`
   readonly JAVAWS_SRC_SRC=`ls $PROJECT_TOP/artifact-all-dependencies/target/icedtea-web-all-dependencies-*.jar | grep sources`
 fi
 


launchers/configure.sh
```diff
@@ -75,7 +75,7 @@ if [ "x$ITW_LIBS" == "xDISTRIBUTION" ] ; then
   readonly JAVAWS_SRC=`ls $PROJECT_TOP/artifact-no-dependencies/target/icedtea-web-no-dependencies-*.jar | grep -v sources | grep -v javadoc`
   readonly JAVAWS_SRC_SRC=`ls $PROJECT_TOP/artifact-no-dependencies/target/icedtea-web-no-dependencies-*.jar | grep sources`
 else
-  readonly JAVAWS_SRC=`ls $PROJECT_TOP/artifact-all-dependencies/target/icedtea-web-all-dependencies-*.jar | grep -v sources | grep -v javadoc`
+  readonly JAVAWS_SRC=`ls $PROJECT_TOP/artifact-all-dependencies/target/icedtea-web-all-dependencies-*-shaded.jar`
```
sclassen: If I check in my target folder maven is currently building no `icedtea-web-all-dependencies-*-shaded.jar`.
Do you see such a file in your target folder??

don-vip: Yes, with these command lines:

```
mvn clean install -DskipTests=true
ls -l artifact-all-dependencies/target/
```
 => no shaded jar

```
mvn install -DskipTests=true -Plaunchers
ls -l artifact-all-dependencies/target/
```
 => shaded jar

```
$ ls -l artifact-all-dependencies/target/
total 11496
-rwxrwx---+ 1 2678467 31 oct.  23:34 icedtea-web-all-dependencies-2.0.0-SNAPSHOT.jar
-rwxrwx---+ 1 2678467 31 oct.  23:34 icedtea-web-all-dependencies-2.0.0-SNAPSHOT-shaded.jar
-rwxrwx---+ 1 1866323 31 oct.  23:34 icedtea-web-all-dependencies-2.0.0-SNAPSHOT-sources.jar
drwxrwx---+ 1       0 31 oct.  23:30 maven-archiver
-rwxrwx---+ 1 2678466 31 oct.  23:34 original-icedtea-web-all-dependencies-2.0.0-SNAPSHOT.jar
-rwxrwx---+ 1 1866322 31 oct.  23:30 original-icedtea-web-all-dependencies-2.0.0-SNAPSHOT-sources.jar

```

sclassen: The "-shaded" jar is only a temporary file created by the maven-shade-plugin.
Once the plugin finishes it will replace the original jar with the temporary "-shaded" jar.

So you should not rely on this shaded file as it will only exist during the execution of the maven-shade-plugin

See https://github.com/apache/maven-shade-plugin/blob/master/src/main/java/org/apache/maven/plugins/shade/mojo/ShadeMojo.java#L514

don-vip: OK I have updated the PR to exclude shaded jar

PR #503, opened 2019-11-08
Support unsigned jars (only by specific config)

hendrikebbers:
ITW already provide the support for unsigned jars if a specific config property is defined and if the security is deactivated. Since the property is only used at 1 place (for this check) it would make sense to depend only on this property to support unsigned jars.

@judovana do you think this is ok or do you see any security issues here?

### DIFF
--- a/core/src/main/java/net/sourceforge/jnlp/runtime/JNLPClassLoader.java
+++ b/core/src/main/java/net/sourceforge/jnlp/runtime/JNLPClassLoader.java
@@ -354,8 +354,7 @@ public class JNLPClassLoader extends URLClassLoader {
     }
 
     private static boolean isCertUnderestimated() {
-        return Boolean.parseBoolean(JNLPRuntime.getConfiguration().getProperty(ConfigurationConstants.KEY_SECURITY_ITW_IGNORECERTISSUES))
-                && !JNLPRuntime.isSecurityEnabled();
+        return Boolean.parseBoolean(JNLPRuntime.getConfiguration().getProperty(ConfigurationConstants.KEY_SECURITY_ITW_IGNORECERTISSUES));
     }
 
     private static void consultCertificateSecurityException(LaunchException ex) throws LaunchException {


hendrikebbers:
I just saw that the check method is called several times. Will have a deeper look but would really like to have the possibility to use unsigned jars by config

judovana:
Hi!
I'm not sure it is the right place and thing to do.
IIRC Those checks are for wrongly signed jars, and for mixed signatures (both different signatures and mix of signed/ussigned jars). Allthose three issues should be handled very carefully.

Simply unsigned jars should ran out of the box.

judovana:
To highlight it, to run signed + unsigned jars with all-permissions or similarly, or to run mixed signatures in same way, IS possible bug hole.

benjaminrein:
A secure way to implement the use of unsigned content is already in the spec. This can make sense if for whatever reason it's not possible to sign the code itself. E.g. unsigned applet or Web Start app is embedded in a 3rd party application und cannot be changed.

The solution for this is using a deployment ruleset, which allows execution of the unsigned code. The deployment ruleset is just an xml file, which is packaged inside a **signed** jar.

More details:
https://docs.oracle.com/javase/8/docs/technotes/guides/deploy/deployment_rules.html

sclassen:
The deployment ruleset is an oracle specific addition to web start.
It is currently not supported by ITW

judovana:
> The deployment ruleset is an oracle specific addition to web start.
> It is currently not supported by ITW

yah. In itw you have to tune your local config or sinle-run cmdline switchis to allow such app to run. TY for not pushing it.

judovana:
> The deployment ruleset is an oracle specific addition to web start.
> It is currently not supported by ITW

unluckily following what proprietary javaws is dong is very often necessary. If people are moving to ITW, they ae asking what differecnes they will face. None is the desired answer, but unluckily it is far frmo true

PR #508, opened 2019-11-13
Fix cache folder with space use case

dimitryc:
The problem is reproducible with cache folders that contain spaces. After starting an application, the resources are downloaded to cache folder but attempt to load main class fails with java.lang.ClassNotFoundException.

The problem seems to be caused by inconsistent escaping/unescaping of values of the map - net.sourceforge.jnlp.runtime.CachedJarFileCallback.mapping (associates remote URL with local URL).  Adding values to this map occurs in CachedJarFileCallback.addMapping(remoteURL, localURL) that is called from JNLPClassLoader.activateJars() which performs escaping (localFile->URI->URL) replacing spaces with %20. The getter is in CachedJarFileCallback.retrieve(URL) - for local files, it creates JarFile using escaped value and fails with FileNotFoundException (and swallows it later).The proposed fix is to unescape values for local files.

### DIFF
--- a/core/src/main/java/net/sourceforge/jnlp/runtime/CachedJarFileCallback.java
+++ b/core/src/main/java/net/sourceforge/jnlp/runtime/CachedJarFileCallback.java
@@ -106,7 +106,7 @@ final class CachedJarFileCallback implements URLJarFileCallBack {
 
         if (UrlUtils.isLocalFile(localUrl)) {
             // if it is known to us, just return the cached file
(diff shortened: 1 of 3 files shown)


judovana:
Hi. the fix is good. Do you mind to add at lest unit test? Ideally also integration test?
Thank you very much! (btw no, is also an answer, but test wolyd be really appreciate, otherwise it can be broken again in future)

dimitryc:
Thanks for the review.
Added new test, the test fails without the fix and passes with the fix.

PR #509, opened 2019-11-13
Launcher catchup 4 - Native Debug in Rust

AurumTheEnd:
This was my first Rust experience, so hopefully the code's at least mediocre :D

### DIFF
--- a/launchers/rust-launcher/src/main.rs
+++ b/launchers/rust-launcher/src/main.rs
@@ -83,6 +83,24 @@ fn is_splash_forbidden_testable(vars: Vec<(String, String)>, aargs: Vec<String>)
     false
 }
 
+fn is_native_debug_enabled(args: &Vec<String>) -> bool {
+    for s in args {
+        if clean_param(s.to_string()) == ("-nativeDebug") {
+            return true;
+        }
+    }
+    return false;
+}
+
+fn get_native_debug_parameter(vars: Vec<(String, String)>, parameter: String) -> Option<String> {
+    for (key, value) in vars {
+        if key == parameter{
+            return Some(value);
+        }
+    }
+    return None;
+}
+
 fn main() {
     let os;
     #[cfg(windows)]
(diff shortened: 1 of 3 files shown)


launchers/rust-launcher/src/main.rs
```diff
@@ -151,6 +169,35 @@ fn compose_arguments(java_dir: &std::path::PathBuf, original_args: &std::vec::Ve
...
+            Some(value) => {
+                port = value;
+            }
+            _none => {
+                port = String::from("9009");
+            }
+        }
+
+        native_debug_toggle = String::from("-Xdebug");
+        native_debug_config = [String::from("-Xrunjdwp:transport=dt_socket,server=y,suspend="), suspend, String::from(",address="), port].concat();
```
HlinaCZ: I would suggest using `format!()` macro.

launchers/rust-launcher/src/main.rs
```diff
@@ -151,6 +169,35 @@ fn compose_arguments(java_dir: &std::path::PathBuf, original_args: &std::vec::Ve
...
+    if native_debug {
+        let (suspend, port);
+        match get_native_debug_parameter(env::vars().collect::<Vec<_>>(), String::from("ITW_NATIVEDEBUG_SUSPEND")){
+            Some(value) => {
+                suspend = value;
+            }
+            _none => {
+                suspend = String::from("n");
+            }
+        }
```
HlinaCZ: In my opinion `if let` would be easier to read.

launchers/rust-launcher/src/main.rs
```diff
@@ -151,6 +169,35 @@ fn compose_arguments(java_dir: &std::path::PathBuf, original_args: &std::vec::Ve
...
+        match get_native_debug_parameter(env::vars().collect::<Vec<_>>(), String::from("ITW_NATIVEDEBUG_PORT")){
+            Some(value) => {
+                port = value;
+            }
+            _none => {
+                port = String::from("9009");
+            }
+        }
```
HlinaCZ: ditoo

launchers/rust-launcher/src/main.rs
HlinaCZ: Btw it is more idiomatic using type `None` when matching value returned from Option instead of default.

launchers/rust-launcher/src/main.rs
```diff
@@ -83,6 +83,24 @@ fn is_splash_forbidden_testable(vars: Vec<(String, String)>, aargs: Vec<String>)
     false
 }
 
+fn is_native_debug_enabled(args: &Vec<String>) -> bool {
+    for s in args {
+        if clean_param(s.to_string()) == ("-nativeDebug") {
```
judovana: please extract nativeDebug to constant. It is already used twice

launchers/rust-launcher/src/main.rs
judovana: yes please!

judovana:
Aurum,  tested also on windows? Otherwise loks good to me

AurumTheEnd:
> Aurum, tested also on windows? Otherwise loks good to me

I am unfortunately unable to build the rust launchers on Windows - it reports a LLVM error when executing https://github.com/AdoptOpenJDK/IcedTea-Web/blob/master/launchers/utils.sh#L41. Trying to find a fix.

judovana:
Really? last time we have spoken it was already on wix setup:(

AurumTheEnd:
> 
> 
> Really? last time we have spoken it was already on wix setup:(

Retraced my steps and it is a fault with me being in a Windows VM and the project being on a shared drive, as is mentioned [here](https://github.com/rust-lang/rust/issues/54216).

AurumTheEnd:
@judovana Copied directory to a local path, fixed an existing typo and it builds OK now.

PR #530, opened 2019-11-27
Sign check empty jnlp

hendrikebbers:
For an empty jnlp (no resources, no extensions) the check was wrong


core/src/main/java/net/sourceforge/jnlp/runtime/JNLPClassLoader.java
```diff
@@ -647,22 +653,28 @@ private boolean shouldFilterInvalidJars() {
      */
     private void initializeResources() throws LaunchException {
 
-        JARDesc[] jars = resources.getJARs();
+        final JARDesc[] jars = resources.getJARs();
 
         if (jars.length == 0) {
```
sclassen: Is this if correct?
I see that it has not changed. I never the less question its correctness.

For me the code now reads:
If the main JNLP has no jars, then do not check for extension JNLPs...

hendrikebbers: As far as I can say a new JNLPClassLoader will be created for each JNLP extension in an JNLP file. Based on this the code will be executed for each JNLP file (parent & Childs/Extensions). @janakmulani has some samples that we maybe can check again.

PR #533, opened 2019-11-29
Extension point

sclassen:


core/src/main/java/net/adoptopenjdk/icedteaweb/extensionpoint/ExtensionPoint.java
```diff
@@ -0,0 +1,41 @@
...
+ */
+public interface ExtensionPoint {
+
+    ExtensionPoint DEFAULT = new ExtensionPoint() {};
+
+    default JvmLauncher getJvmLauncher() {
+        return new ItwJvmLauncher();
+    }
+
+    default MenuAndDesktopIntegration getMenuAndDesktopIntegration(final DeploymentConfiguration configuration) {
```
hendrikebbers: rename methods to "create*"

PR #565, opened 2020-01-09
fix for failed RegistryQueryTest on Windows

lherschi:


core/src/test/java/net/adoptopenjdk/icedteaweb/proxy/ie/RegistryQueryTest.java
```diff
@@ -15,7 +15,7 @@ public void getRegistryValuesFromLines() throws Exception{
         //Given
         final String key = "HKEY_CURRENT_USER\\Software\\Microsoft\\Windows\\CurrentVersion\\Internet Settings";
         final String content = IOUtils.readContentAsUtf8String(RegistryQueryTest.class.getResourceAsStream("reg_query.out"));
-        final List<String> lines = Arrays.asList(content.split(System.lineSeparator()));
+        final List<String> lines = Arrays.asList(content.split("\n"));
```
sclassen: Could you use `\\R` for splitting?
see: https://stackoverflow.com/a/31060125

PR #575, opened 2020-01-28
Fixes and code cleanup

sclassen:


core/src/main/java/net/sourceforge/jnlp/util/logging/OutputController.java
```diff
@@ -300,10 +300,10 @@ private SingleStreamLogger getAppFileLog() {
         private static final SingleStreamLogger INSTANCE = initSystemLogger();
 
         private static SingleStreamLogger initSystemLogger() {
-            if (OsUtil.isWindows()) {
-                return new WinSystemLog();
-            } else {
+            if (OsUtil.isLinux()) {
```
sclassen: @hendrikebbers 
This changes the behavior for MacOS as before it was handed the UnixSystemLog and now it gets the MoopSystemLog. Is this change intentionally??

PR #583, opened 2020-02-05
Fix hanging local host lookup

AndreasEhret:
- provide alternative localhost detection to avoid thread hangs within the InetAddress
- using 3rd party lib that unlike java.net.InetAddress.getByName() will not trigger a DNS lookup for an invalid address or host


sclassen:
@hendrikebbers 
Should we try to implement the IPv6 support ourselves or do we want to shade this as a dependency?

AndreasEhret:
added shading of ipaddress library.

PR #593, opened 2020-02-21
Default Value for Locked Prop if prop is not specified

janakmulani:


sclassen:
Closing this as we decided to follow an other approach

PR #594, opened 2020-02-21
Locked prop def value

janakmulani:
We need to take care of both the following cases:
1. If locked prop is specified without real property then create real property with default value
2. If a prop with def value is locked then save it.


sclassen:
extracted relevant changes into https://github.com/AdoptOpenJDK/IcedTea-Web/commit/07df88cb35662b4dba816a2aa4eb856a6575c560

PR #596, opened 2020-02-25
sun.awt.DisableMixing

janakmulani:


janakmulani:
@AndreasEhret pl merge now

core/src/main/java/net/adoptopenjdk/icedteaweb/jnlp/element/security/SecurityDesc.java
```diff
@@ -232,6 +233,7 @@
             new PropertyPermission(HTTP_KEEP_ALIVE, PROPERTY_RW_ACTION),
             new PropertyPermission(AWT_SYNC_LWREQUESTS, PROPERTY_RW_ACTION),
             new PropertyPermission(AWT_WINDOW_LOCATION_BY_PLATFORM, PROPERTY_RW_ACTION),
+            new PropertyPermission(AWT_DISABLE_MIXING, PROPERTY_RW_ACTION),
```
sclassen: The set of properties access is defined in JSR-56.
I don't know if we should change this

artifact-all-dependencies/pom.xml
```diff
@@ -40,7 +40,6 @@
                         </goals>
                         <configuration>
                             <createSourcesJar>true</createSourcesJar>
-                            <shadeSourcesContent>true</shadeSourcesContent>
```
sclassen: How is this related to this PR?

PR #597, opened 2020-02-28
encoding of shortcutlist.txt file

janakmulani:


core/src/main/java/net/sourceforge/jnlp/util/WindowsDesktopEntry.java
```diff
@@ -151,9 +154,17 @@ private void manageShortcutList(ManageMode mode, String path) throws IOException
...
         if (ManageMode.A == mode) {
-            List<String> lines = Files.readAllLines(getWindowsShortcutsFile().toPath(), UTF_8);
+            List<String> lines = null;
+            // if UTF-8 fails, try ISO-8859-1
+            try {
+                LOG.debug("Reading Shortcuts with UTF-8");
+                lines = Files.readAllLines(getWindowsShortcutsFile().toPath(), UTF_8);
+            } catch (MalformedInputException me) {
+                LOG.debug("Reading Shortcuts with ISO-8859-1");
+                lines = Files.readAllLines(getWindowsShortcutsFile().toPath(), ISO_8859_1);
```
sclassen: ISO-8895-1 is a subset of UTF-8 thos if the file is encoded in ISO-8859-1 it can be paresd using UTF-8.
Therefote it makes no sense to fall back to ISO-8859-1.

core/src/main/java/net/sourceforge/jnlp/util/WindowsDesktopEntry.java
```diff
@@ -171,10 +182,13 @@ private void manageShortcutList(ManageMode mode, String path) throws IOException
                 }
             }
             if (fAdd) {
-                LOG.debug("Adding Shortcut to list = ", sItem);
-                String scInfo = file.getFileLocation().toString() + ",";
-                scInfo += path + "\r\n";
-                Files.write(getWindowsShortcutsFile().toPath(), scInfo.getBytes(), StandardOpenOption.APPEND);
+                LOG.debug("Adding Shortcut to list = {}", sItem);
+                LOG.debug("Default encoding is {}", Charset.defaultCharset().name());
+                StringBuilder stringBuilder = new StringBuilder().append(file.getFileLocation().toString())
```
sclassen: With modern java versions there is no longer a performance gain of StringBuilder over simple string concatenation.

core/src/main/java/net/sourceforge/jnlp/util/WindowsDesktopEntry.java
janakmulani: I have tested this the char that caused problem with UTF8 worked with ISO

sclassen: Appologies Janak. Only the first 128 characters of ISO-8859-1 are a subset of UTF-8.
Therefore yes some characters are different.
But if the file was written in CP-1252, wouldn't it make more sense to fall back to this encoding?

sclassen: Or maybe fall back to the default encoding of the machine. As the file was most likely previously written using the default encoding.

janakmulani: Yes default makes more sense. On Windows default is windows-1252. Will make some tests and change.

core/src/main/java/net/sourceforge/jnlp/util/WindowsDesktopEntry.java
janakmulani: changed and committed

core/src/main/java/net/sourceforge/jnlp/util/WindowsDesktopEntry.java
janakmulani: Changed and committed

PR #598, opened 2020-03-02
Improve shortcut management

sclassen:


core/src/main/java/net/sourceforge/jnlp/util/WindowsDesktopEntry.java
```diff
@@ -145,51 +143,44 @@ public void createWindowsMenu() throws IOException {
...
+            LOG.debug("Fallback to reading Shortcuts with default encoding {}", Charset.defaultCharset().name());
+            return Files.readAllLines(shortcutFile.toPath(), Charset.defaultCharset());
+        }
+    }
+
+    private boolean needToAddNewShortcutEntry(String path, List<String> lines) {
+        return lines.stream()
+                .map(line -> line.split(","))
+                .map(array -> array[1])
+                .anyMatch(sPath -> sPath.equalsIgnoreCase(path));
```
janakmulani: noneMatch - if we don't find a match then we need to add

sclassen: thanks, fixed

core/src/main/java/net/sourceforge/jnlp/util/WindowsDesktopEntry.java
```diff
@@ -145,51 +143,44 @@ public void createWindowsMenu() throws IOException {
...
-            }
-            if (fAdd) {
-                LOG.debug("Default encoding is {}", Charset.defaultCharset().name());
-                LOG.debug("Adding Shortcut to list = {} with UTF-8 encoding", sItem);
-                String scInfo = file.getFileLocation().toString() + ",";
-                scInfo += path + "\r\n";
-                Files.write(getWindowsShortcutsFile().toPath(), scInfo.getBytes(UTF_8), StandardOpenOption.APPEND);
-            }
+        LOG.debug("Using WindowsShortCutManager {}", shortcutFile);
+        final List<String> lines = readAllLine(shortcutFile);
```
AndreasEhret: wouldn't be "readAllLines" a better method name here?

sclassen: full ack

PR #614, opened 2020-04-01
process termination

janakmulani:


core/src/main/java/net/sourceforge/jnlp/runtime/Boot.java
```diff
@@ -110,9 +110,8 @@
     public static void main(String[] args) {
         EnvironmentPrinter.logEnvironment(args);
         final int status = mainWithReturnCode(args);
-        if (status != 0) {
-            System.exit(status);
-        }
+        LOG.debug("Exiting Boot.main().");
```
sclassen, line 113: Logging is closed in method `mainWithReturnCode(args)`

PR #615, opened 2020-04-02
do not use logging framework within the logging

sclassen:


core/src/main/java/net/sourceforge/jnlp/util/logging/OutputController.java
```diff
@@ -88,7 +91,13 @@ public void run() {
                     }
 
                 } catch (Throwable t) {
-                    LOG.error(IcedTeaWebConstants.DEFAULT_ERROR_MESSAGE, t);
+                    final File logDir = PathsAndFiles.LOG_DIR.getFile();
+                    final File file = new File(logDir, "log-trace-" + System.currentTimeMillis() + ".log");
```
AndreasEhret, line 95: I would leave a brief comment here why we do this.

PR #620, opened 2020-04-08
java module args

janakmulani:


common/src/test/java/net/adoptopenjdk/icedteaweb/jvm/JvmUtilsTest.java
```diff
@@ -90,4 +96,114 @@ public void testJava8JavaVMArgs() {
...
+        } catch (IllegalArgumentException ile) {
+            fail(ile.getMessage());
+        }
+
+        final String java_vm_bad_args = "-Dsun.java2d.d3d=false -Dsun.java2d.dpiaware=false  --add-opens=java.desktop/sun.print=ALL-UNNAMED --list-modules --add-exports=java.desktop/sun.print=ALL-UNNAMED --add-exports=java.desktop/sun.swing=ALL-UNNAMED --add-exports=java.desktop/sun.swing.table=ALL-UNNAMED --add-exports=java.desktop/sun.swing.plaf.synth=ALL-UNNAMED --add-opens=java.desktop/javax.swing.plaf.synth=ALL-UNNAMED --add-opens=java.desktop/javax.swing.plaf.basic=ALL-UNNAMED --add-opens=java.desktop/javax.swing=ALL-UNNAMED --add-opens=java.desktop/javax.swing.tree=ALL-UNNAMED --add-opens=java.desktop/java.awt.event=ALL-UNNAMED --add-exports=java.desktop/com.sun.java.swing.plaf.windows=ALL-UNNAMED --add-exports=java.desktop/sun.awt.shell=ALL-UNNAMED --add-exports=java.desktop/com.sun.awt=ALL-UNNAMED --add-exports=java.base/sun.security.action=ALL-UNNAMED";
+        try {
+            JvmUtils.checkVMArgs(java_vm_bad_args);
+            fail("Illegal arg accepted!");
+        } catch (IllegalArgumentException ile) {
+            assertTrue(true);
```
sclassen, line 121: a comment would be better than `assertTrue(true)` which is always true...

common/src/test/java/net/adoptopenjdk/icedteaweb/jvm/JvmUtilsTest.java
```diff
@@ -90,4 +96,114 @@ public void testJava8JavaVMArgs() {
...
+    }
+
+    @Test
+    public void testMergeJavaModuleVMArgs() {
+
+        final String[] usrStrArr = new String[] {"-DnoModuleArg=bbb", "--add-reads=java.base=ALL-UNNAMED,java.xxx"};
+        final List<String> usrDefArgs = new ArrayList(Arrays.asList(usrStrArr));
+
+        final List<String> result = JvmUtils.mergeJavaModulesVMArgs(usrDefArgs);
+        assertTrue(result.contains("-DnoModuleArg=bbb"));
```
sclassen, line 142: can you please add the following assertion:

`assertEquals(JvmUtils.getPredefinedJavaModulesVMArgumentsMap().size() + 1, result.size());`

common/src/test/java/net/adoptopenjdk/icedteaweb/jvm/JvmUtilsTest.java
```diff
@@ -90,4 +96,114 @@ public void testJava8JavaVMArgs() {
...
+    }
+
+    @Test
+    public void testMergeDuplicateJavaModuleVMArgs() {
+
+        final String[] usrStrArr = new String[] {"-DnoModuleArg=bbb", "--add-reads=java.base=ALL-UNNAMED,java.xxx", "--add-reads=java.base=ALL-UNNAMED,java.yyy"};
+        final List<String> usrDefArgs = new ArrayList(Arrays.asList(usrStrArr));
+
+        final List<String> result = JvmUtils.mergeJavaModulesVMArgs(usrDefArgs);
+        assertTrue(result.contains("-DnoModuleArg=bbb"));
```
sclassen, line 153: can you also assert in this test the size of `result`

common/src/test/java/net/adoptopenjdk/icedteaweb/jvm/JvmUtilsTest.java
```diff
@@ -90,4 +96,114 @@ public void testJava8JavaVMArgs() {
             fail(ile.getMessage());
         }
     }
+
+    @Test
+    public void testJava9JavaVMArgs() {
```
sclassen, line 101: can you please split this test such that only one call to `JvmUtils.checkVMArgs()` happens in a test method

common/src/test/java/net/adoptopenjdk/icedteaweb/jvm/JvmUtilsTest.java
```diff
@@ -90,4 +96,114 @@ public void testJava8JavaVMArgs() {
...
+        assertTrue(result.contains("-DnoModuleArg=bbb"));
+        assertTrue(result.contains("--add-reads=java.base=ALL-UNNAMED,java.desktop,java.yyy"));
+        assertTrue(result.contains("--module-path=java.base=java.xxx"));
+        assertTrue(result.contains("--add-opens=java.base=java.aaa"));
+        assertTrue(result.contains("--add-modules=java.base=java.bbb"));
+        assertTrue(result.contains("--patch-module=java.base=java.ccc"));
+    }
+
+    @Test
+    public void testValidJavaModuleVMArgs() {
```
sclassen, line 172: please split this test

common/src/test/java/net/adoptopenjdk/icedteaweb/jvm/JvmUtilsTest.java
```diff
@@ -90,4 +96,114 @@ public void testJava8JavaVMArgs() {
...
+    @Test
+    public void testInvaldiJavaModuleVMArgs() {
+
+        final String usrArgs = "-Dsun.java2d.d3d=true --list-module=java.base=ALL-UNNAMED,java.xxx --add-reads=java.base=ALL-UNNAMED,java.yyy";
+
+        try {
+            JvmUtils.checkVMArgs(usrArgs);
+            fail("Illegal arg accepted!");
+        } catch (IllegalArgumentException ile) {
+            assertTrue(true);
```
sclassen, line 196: replace with a comment or use `@Test(expected = IllegalArgumentException.class)` to have jUnit make sure the expected exception is thrown.

common/src/main/java/net/adoptopenjdk/icedteaweb/jvm/JvmUtils.java
```diff
@@ -301,9 +317,83 @@ public static String findOnPath(final String[] bins) {
...
                 }
-
             }
         }
         return exec;
     }
+
+    /**
+     * https://docs.oracle.com/javase/9/tools/java.htm#JSWOR624
+     * Java 9 VM args. specified by app developer in jnlp file.
```
AndreasEhret, line 327: This javaDoc should be updated/improved (if the method remains) as it still references java9 instead of java modules and the itw-modularjdk file

common/src/main/java/net/adoptopenjdk/icedteaweb/jvm/JvmUtils.java
```diff
@@ -301,9 +317,83 @@ public static String findOnPath(final String[] bins) {
...
+     * ITW add its own set of jvm args (specified in file itw-modularjdk.args) required for its own execution
+     *
+     * @return
+     */
+    public static String[] getValidStartingJavaModuleVMArguments() {
```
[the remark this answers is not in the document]
janakmulani: yes. but I just followed the pattern like for other properties. perhaps the idea  is avoid method call for each vm arg and instead use preinitialized data structure.

common/src/test/java/net/adoptopenjdk/icedteaweb/jvm/JvmUtilsTest.java
janakmulani: Right

common/src/main/java/net/adoptopenjdk/icedteaweb/jvm/JvmUtils.java
janakmulani: ok

PR #626, opened 2020-04-17
change file store impl to make json more compact

sclassen:
The new JSON file has the following structure:

```
{
  "version": "1",
  "userDecisions": [
    {
      "domain": "http://localhost",
      "jarNames": [
        "eager.jar",
        "lazy.jar",
        "native.jar"
      ],
      "decisions": {
        "RUN_UNSIGNED_APPLICATION": "DENY",
        "RUN_MISSING_ALAC_APPLICATION": "ALLOW",
        "RUN_MISSING_PERMISSIONS_APPLICATION": "SANDBOX"
      }
    }
  ]
}
```

The advantage is that multiple decisions can be stored in the same entry.

### DIFF
--- a/core/src/main/java/net/adoptopenjdk/icedteaweb/userdecision/FileStoreEntry.java
+++ b/core/src/main/java/net/adoptopenjdk/icedteaweb/userdecision/FileStoreEntry.java
@@ -1,54 +1,34 @@
 package net.adoptopenjdk.icedteaweb.userdecision;
 
-import net.adoptopenjdk.icedteaweb.Assert;
-
 import java.net.URL;
-import java.util.Objects;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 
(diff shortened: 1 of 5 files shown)


core/src/main/java/net/adoptopenjdk/icedteaweb/userdecision/FileStoreEntry.java
```diff
@@ -1,54 +1,34 @@
...
-public class FileStoreEntry {
-    private UserDecision.Key userDecisionKey;
-    private String userDecisionValue;
-
-    private URL domain;
-    private Set<String> jarNames;
+class FileStoreEntry {
+    private final URL domain;
+    private final Set<String> jarNames;
+    private final Map<UserDecision.Key, String> decisions = new HashMap<>();
```
sclassen, line 12: This is the main change.
A `FileStoreEntry` has now a map of `UserDecision.Key -> String`

PR #632, opened 2020-04-28
Multiple eager parts

sclassen:
A JNLP file may have more than one eager part.
All the jars from all eager parts must be validated together.

This change ensures the above. The old implementation validated each eager part by itself.


hendrikebbers:
@sclassen since this is a PR to the classloader-prototype branch I would just merge it :)

sclassen:
let's wait a few more days to see if we get more feedback from Italy about the lazy loading

PR #636, opened 2020-05-07
repaint manager removed

hendrikebbers:
Old RepaintManager removed that was never used


sclassen:
What did you do with the swing utils?

AndreasEhret:
Could you elaborate about the reasons for the change to give reviewers a change to follow your motivation for doing this?

hendrikebbers:
Looks like I have some Git Problems with line separators in commits or the file has a wrong line separator. In the SwingUtil class only 3 lines should have changed. You can see this by ignoring Whitespace in the diff. I can understand that you have no idea what happened here when seeing the diff without ignoring white spaces.

About the change: 
the repaint manager has an unknown license. This will end in a general license issue for ITW. Since the repaint manager is never used (debug is never true). I just removed the class.

judovana:
Repaint manager in JDK was (is?) broken for some time. IIRC, that was jdk11 specific issue.  That was making ITW quite unstable. Ths class for here, to help debug the root cause. You had to manually set  it to true, and then observe waht is happening.

PR #643, opened 2020-05-22
Replace xml-parser with default java dom parser

hendrikebbers:
Replaced custom XML parser with the default Java DOM parser. This makes the code better maintainable and removes the bad licence of the customer parser that was copied from another project.

Looks like I have a problem with commits since modified files line separators are all CRLF now. Maybe @sclassen can help me here :) For now please ignore whitespaces for the review. I will try to add a commit with the correct (old) line separators next week.

Additional points:
- @Ignore some tests the tested malformed xml that is not valid and therefore can not be parsed with the default Java DOM parser
- Needed to @Ignore ParserCornerCasesTest.testCommentInElements2 but I do not understand the result of the test ("+1.0" instead of "1.0") - @sclassen @AndreasEhret any idea?


sclassen:
@hendrikebbers 
I cleaned up the line endings and did some other minor changes.
Notably I use `Reader` instead of `InputStream` where possible. Because the Reader uses a Character source and thus does not convert the string to bytes and back which always has the potential of encoding problems.

I have some more comments but GitHub is currently unstable.
I will try to add my review later

sclassen:
@hendrikebbers I renamed your original branch to `original-new-xml-parser`

PR #644, opened 2020-05-22
Licence header in same format

hendrikebbers:
Same GPLv2, LGPL & GPLv2+CE licence format in all files.
This PR does not change or modify any licence. It only changes the format of all licences.

Still have my git problem and CRLF is used as line separator :(


sclassen:
fixe line endings
Some classes do not have a license header
unfortunately GitHub is unreliable and I cannot add comments to the diff

hendrikebbers:
@sclassen "Some classes do not have a license header" -> I have not added new licence headers in this commit. Or have you seen files where the licence header was removed?

Issue #645, opened 2020-05-25
SingleInstanceService works correctly just for the first application run


vnashkev:
After first start lock file with a port number is created 

C:\Users\user\AppData\Local\Temp\user\netx\locks\http___localhost_jnlp.jnlp.

Lock file still exists after application termination. 
Next time SingleInstanceServer fails to start with the following exception:
```
java.io.IOException: Cannot rename C:\Users\user\AppData\Local\Temp\user\netx\locks\http___localhost_jnlp.jnlp.temp to C:\Users\user\AppData\Local\Temp\user\netx\locks\http___localhost_jnlp.jnlp
	at net.adoptopenjdk.icedteaweb.io.FileUtils.createRestrictedFile(FileUtils.java:331)
	at net.adoptopenjdk.icedteaweb.io.FileUtils.createRestrictedFile(FileUtils.java:172)
	at net.sourceforge.jnlp.services.SingleInstanceLock.createWithPort(SingleInstanceLock.java:70)
	at net.sourceforge.jnlp.services.SingleInstanceLockTest.testCreateWithPort(SingleInstanceLockTest.java:34)
```

Test  case demonstraits the problem:
```
public final class SingleInstanceLockTest
	extends NoStdOutErrTest
{
    @Test
    public void testCreateWithPort() 
    	throws Exception 
    {
    	final URL url = this.getClass().getClassLoader().getResource("net/sourceforge/jnlp/basic.jnlp");
        final JNLPFile jnlpFile = new JNLPFile(url.openStream(), url, new ParserSettings(false,false,false));
        
        assert jnlpFile!=null;
        
        final SingleInstanceLock sil = new SingleInstanceLock(jnlpFile);
        
        sil.lockFile.delete();
        assert sil.lockFile.exists()==false;
        
        sil.createWithPort(123);
        assert sil.lockFile.exists()==true;
        assert "123".equals(FileUtils.loadFileAsUtf8String(sil.lockFile).trim());
        

        sil.createWithPort(456);
        assert sil.lockFile.exists()==true;
        assert "456".equals(FileUtils.loadFileAsUtf8String(sil.lockFile).trim());
    }
}
```

janakmulani:
Thanks for reporting this issue.

I could reproduce your issue. We will fix it in a future release.

vnashkev:
I have created pull request already :
[https://github.com/AdoptOpenJDK/IcedTea-Web/pull/646](url)

[...]

PR #646, opened 2020-05-25
lock file for SingleInstanceService may be creaeted even if old one s…

vnashkev:
…till exists.

Fix for https://github.com/AdoptOpenJDK/IcedTea-Web/issues/645

### DIFF
--- a/common/src/main/java/net/adoptopenjdk/icedteaweb/io/FileUtils.java
+++ b/common/src/main/java/net/adoptopenjdk/icedteaweb/io/FileUtils.java
@@ -240,6 +240,8 @@ public final class FileUtils {
     private static void createRestrictedFile(File file, boolean isDir) throws IOException {
 
         File tempFile = new File(file.getCanonicalPath() + ".temp");
+        FileUtils.deleteWithErrMesg(tempFile, "Could not delete [" + tempFile + "]");
+
 
         if (isDir) {
             if (!tempFile.mkdir()) {
--- a/core/src/test/java/net/sourceforge/jnlp/services/SingleInstanceLockTest.java
+++ b/core/src/test/java/net/sourceforge/jnlp/services/SingleInstanceLockTest.java
@@ -1,0 +1,38 @@
+package net.sourceforge.jnlp.services;
+
+import java.net.URL;
+
+import org.junit.Test;
+
+import net.adoptopenjdk.icedteaweb.io.FileUtils;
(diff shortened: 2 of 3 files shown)


core/src/test/java/net/sourceforge/jnlp/services/SingleInstanceLockTest.java
```diff
@@ -0,0 +1,38 @@
...
+        
+        sil.lockFile.delete();
+        assert sil.lockFile.exists()==false;
+        
+        sil.createWithPort(123);
+        assert sil.lockFile.exists()==true;
+        assert "123".equals(FileUtils.loadFileAsUtf8String(sil.lockFile).trim());
+        
+
+        sil.createWithPort(456);
```
hendrikebbers, line 34: Should this really be possible? I would assume that a lock must be removed before a new one can be created.

vnashkev: Icedtea-web does not remove lock file neither before application start nor after application termination. At least i was not able to find this is source code.

hendrikebbers: mmh, ok. I would assume that creating a lock will create a file and remove a lock will delete the file. Creating a without deleting a previous lock should throw an exception. But I will have a deeper look.

sclassen: The naming of this is bad.
The lock file is not used for locking but for broadcasting a port.

This port is then used as
- a lock because only one process can listen on a port
- a communication channel to inform the running application that a second instance should be started

common/src/main/java/net/adoptopenjdk/icedteaweb/io/FileUtils.java
```diff
@@ -240,6 +240,8 @@ public static void deleteWithErrMesg(File f) {
     private static void createRestrictedFile(File file, boolean isDir) throws IOException {
 
         File tempFile = new File(file.getCanonicalPath() + ".temp");
+        FileUtils.deleteWithErrMesg(tempFile, "Could not delete [" + tempFile + "]");
```
sclassen, line 243: maybe it would be better to do this at the end of the method to prevent race conditions where two threads try to create the same file

PR #650, opened 2020-06-02
default launchers made aware of IPAddress dependence

judovana:


hendrikebbers:
@judovana if the build does not use the jar-with-dependencies where does it get all the 3rdparty jars? I found `SYSTEM_JARS=/usr/share` in the script. Does it assume to have the jars locally in this folder?

judovana:
hi!
There are two jars with dependecies, which @sclassen  implemented with a genius loci.
One have everything - all itw classes, and all depndencies. This one is used for portable builds.
Second one have only all itw classes without depndencies. This one is for distributions, and the  missing jars are composed on runtime. The search is quite limited, as -the lcoations are moreover preset during build.

Thanx for asking!
  J.

judovana:
portable builds/distribution builds  - both from scope of default launchers.

PR #653, opened 2020-06-04
Update RELEASE.md configure installer release job

douph1:
add how to integrated last ITW into AdopOpenJDK installer


douph1:
ICEDTEA_WEB_VERSION is a build param of the job but I have not found any value in openjdk-build repo. So I think we always use the default value configured in the job. Correct me if I'm wrong.

Maybe we can set it into pipelines/build/common/openjdk_build_pipeline.groovy ..

hendrikebbers:
THX @douph1

PR #659, opened 2020-06-08
remember original system out in SystemOutLogger

sclassen:


common/src/main/java/net/adoptopenjdk/icedteaweb/logging/LoggerFactory.java
```diff
@@ -6,14 +6,28 @@
 public class LoggerFactory {
 
     private static final String FACTORY_CLASS = "net.sourceforge.jnlp.util.logging.OutputControllerLoggerFactory";
-    private static final LoggerFactoryImpl factory = initFactory();
+    private static final LoggerFactoryImpl factory;
 
-    private static LoggerFactoryImpl initFactory() {
```
hendrikebbers, line 11: Why is this not a method anymore?

common/src/main/java/net/adoptopenjdk/icedteaweb/logging/LoggerFactory.java
```diff
@@ -6,14 +6,28 @@
...
+            ex = e;
+            loggerFactory = new SystemOutLoggerFactory();
+        }
+        factory = loggerFactory;
+
+        // one can only get a logger after the factory has been set.
+        // therefore we postpone all the logging to after this comment
+
+        final Logger LOG = factory.getLogger(LoggerFactory.class);
+        if (ex != null) {
```
hendrikebbers, line 27: if / else can be inlined in the try-catch

common/src/main/java/net/adoptopenjdk/icedteaweb/logging/SystemOutLoggerFactory.java
```diff
@@ -11,6 +12,16 @@
  */
 class SystemOutLoggerFactory implements LoggerFactoryImpl {
 
+    private static final PrintStream out;
+
+    static {
+        final PrintStream tmp = System.out;
+        if (tmp.getClass().getSimpleName().contains("TeeOutputStream")) {
```
hendrikebbers, line 19: really? can we check if the class implements an interface for example or is assignable by another class?

common/src/main/java/net/adoptopenjdk/icedteaweb/logging/LoggerFactory.java
sclassen: because we cannot call LOG.xxx before we assigned the static field `factory`
So if we make this a method which returns the factory we cannot log in that method.

common/src/main/java/net/adoptopenjdk/icedteaweb/logging/LoggerFactory.java
sclassen: tried this and failed. Reason is we cannot log before the static field `factory` is assigned. And assigning it in the try catch block does not compile.

common/src/main/java/net/adoptopenjdk/icedteaweb/logging/SystemOutLoggerFactory.java
sclassen: sorry only the TeeOutputStream poses a problem. And it is not visible from here (different JAR)

PR #660, opened 2020-06-11
Cleanup logging

sclassen:


core/src/main/java/net/sourceforge/jnlp/util/logging/DummyLogger.java
```diff
@@ -0,0 +1,46 @@
+/*Copyright (C) 2013 Red Hat, Inc.
```
hendrikebbers, line 1: Can you remove the licence header in this new file. At least the copyright is not correct.

core/src/main/java/net/sourceforge/jnlp/util/logging/StdInOutErrController.java
```diff
@@ -0,0 +1,89 @@
+/*Copyright (C) 2013 Red Hat, Inc.
```
hendrikebbers, line 1: Please remove the licence header

core/src/main/java/net/sourceforge/jnlp/util/logging/StdInOutErrController.java
```diff
@@ -0,0 +1,89 @@
...
+        return outLog.getStream();
+    }
+
+    public PrintStreamLogger getErr() {
+        return errLog;
+    }
+
+    public synchronized String readLine() throws IOException {
+        if (br == null) {
+            br = new BufferedReader(new InputStreamReader(System.in));
```
hendrikebbers, line 85: I ask myself if we need to store this in a field of the class.

core/src/main/java/net/sourceforge/jnlp/util/logging/UnixSystemLog.java
```diff
@@ -33,43 +33,54 @@
...
         try {
-            String[] ss = s.split("\\n"); //exceptions have many lines
+            CompletableFuture<Void> f = CompletableFuture.completedFuture(null);
             for (String m : ss) {
-                m = m.replaceAll("\t", "    ");
-                ProcessBuilder pb = new ProcessBuilder("logger", "-p","user.err", "--", m);
-                Process p = pb.start();
-                ProcessUtils.waitForSafely(p);
-                LOG.debug("System logger called with result of {}", p.exitValue());
+                f = f.thenRunAsync(() -> logToSyslog(m), executor);
```
hendrikebbers, line 63: As far as I can see f is not used after the call. Therefore I assume that we could create a deamon based SingleThreadExecutor and simple call execute on that one instead of wrapping all in a future.

core/src/main/java/net/sourceforge/jnlp/util/logging/UnixSystemLog.java
```diff
@@ -33,43 +33,54 @@
...
             }
         } catch (Exception ex) {
-            LOG.error(IcedTeaWebConstants.DEFAULT_ERROR_MESSAGE, ex);
+            LOG.error("Error while sending message to Unix system log", ex);
+        }
+    }
+
+    private void logToSyslog(String message) {
+        try {
+            message = message.replaceAll("\t", "    ");
```
hendrikebbers, line 72: lets create a new var and make the method param final :)

core/src/test/java/net/sourceforge/jnlp/DefaultLaunchHandlerTest.java
```diff
@@ -33,113 +33,87 @@
...
-        }
-
-        String getStream1() {
-            return ((AccessibleStream) (super.getOut())).getOut().toString();
-        }
-
-        String getStream2() {
-            return ((AccessibleStream) (super.getErr())).getOut().toString();
-        }
+    @Before
```
hendrikebbers, line 57: I ask myself if we should have an @After to reset the behaviour for tests that will be executed after the tests of this class

core/src/main/java/net/sourceforge/jnlp/util/logging/DummyLogger.java
sclassen: This is a class extracted from another class. So technically the copyright is still with redhat

core/src/main/java/net/sourceforge/jnlp/util/logging/StdInOutErrController.java
sclassen: The code is copied. I guess the field is an optimization to prevent creating the Reader for every call of the method.
But since this is only used in headless mode it has limited effect...

core/src/main/java/net/sourceforge/jnlp/util/logging/UnixSystemLog.java
sclassen: I admit this code is very tricky. It could most likely bee improved.
Since this line is in a for loop `f` is used in the next iteration of the loop.
This makes sure that the next call to `logToSyslog` is only started when the previous has finished.

Making this all run in daemon threads is a good idea. I will check this.

sclassen: Oh, I just see that I already define an executor which will only use daemon threads

PR #665, opened 2020-07-01
Create Messages_tr.properties

oltulu:


sclassen:
@oltulu Thank you for your contribution.

Just one question.
As ITW is still developed the translations will change. None of the current committer are able to update the Turkish translations.
Would you be willing to periodically take that task and update the translation file?

oltulu:
I try to help

PR #670, opened 2020-07-08
DownloadIndicator by ExtensionPoint

hendrikebbers:


core/src/main/java/net/sourceforge/jnlp/runtime/JNLPRuntime.java
```diff
@@ -242,8 +241,9 @@ public static void initialize() throws IllegalStateException {
         System.setProperty("javawebstart.version", "javaws-" +
                 JavaSystemProperties.getJavaVersion());
 
-        if (!isHeadless() && indicator == null)
-            indicator = new DefaultDownloadIndicator();
+        if (!isHeadless()) {
```
sclassen, line 244: Did you drop the null check on purpose?

hendrikebbers: It is always null...

PR #672, opened 2020-07-09
New logging

sclassen:


core/src/main/java/net/adoptopenjdk/icedteaweb/client/commandline/CommandLine.java
```diff
@@ -73,6 +74,11 @@
  */
 public class CommandLine {
 
+    static {
+        // this is placed here above the any thing else to ensure no logger has been created prior to this line
+        FileLog.setLogFileNamePostfix("settings");
```
hendrikebbers, line 79: Why does the CommandLine has the settings prefix?

core/src/main/java/net/adoptopenjdk/icedteaweb/client/console/ConsoleOutputPaneModel.java
```diff
@@ -164,8 +164,8 @@ String createLine(MessageWithHeader m) {
         if (showHeaders) {
             sb.append(m.getHeader().toString(showUser,
                     showOrigin,
-                    showLevel,
                     showDate,
+                    showLevel,
                     showCode,
                     showThread1,
```
hendrikebbers, line 170: What is the difference between thread1 and thread2 ?

core/src/main/java/net/sourceforge/jnlp/util/logging/FileLog.java
```diff
@@ -50,26 +51,48 @@
...
+            logFileNamePrefix = envPrefix;
+        } else {
+            logFileNamePrefix = FILE_LOG_NAME_FORMATTER.format(new Date());
+        }
+
+        final String envPostfix = System.getenv(LOG_POSTFIX_ENV);
+        if (!StringUtils.isBlank(envPostfix)) {
+            logFileNamePostfix = envPostfix;
+        } else {
+            logFileNamePostfix = "itw";
```
hendrikebbers, line 76: we already log in an itw folder. Do we really need as a prefix, too?

core/src/main/java/net/sourceforge/jnlp/util/logging/OutputControllerLevel.java
```diff
@@ -12,11 +12,17 @@
 
     ;
 
+    private final String displayName = (name() + "   ").substring(0,5);
```
hendrikebbers, line 15: Can you do this in the constructor?

core/src/main/java/net/adoptopenjdk/icedteaweb/client/commandline/CommandLine.java
sclassen: the `main()` in CommandLine is used by ITW to start the settings UI.
This change marks the logs from running the settings by appending "settings" to the log file name

core/src/main/java/net/adoptopenjdk/icedteaweb/client/console/ConsoleOutputPaneModel.java
sclassen: one is the hash and other the name of the thread.
I renamed the vars in most locations. will also do it here

core/src/main/java/net/sourceforge/jnlp/util/logging/FileLog.java
sclassen: what else would you do as a postfix???

core/src/main/java/net/sourceforge/jnlp/util/logging/OutputControllerLevel.java
sclassen: what would be the difference???

sclassen:
I merge this for now.
Feel free to improve

PR #674, opened 2020-07-14
wildcards in whitelist - first cut

janakmulani:


core/src/test/java/net/sourceforge/jnlp/util/UrlWhiteListUtilsTest.java
```diff
@@ -0,0 +1,151 @@
+/*
+   Copyright (C) 2012 Red Hat, Inc.
```
hendrikebbers, line 2: is this a new file? in such case pls remove licence header

core/src/test/java/net/sourceforge/jnlp/util/UrlWhiteListUtilsTest.java
```diff
@@ -0,0 +1,151 @@
...
+
+        Assert.assertEquals("http://*", UrlWhiteListUtils.expandedWhiteListUrlString("http://*:*"));
+        Assert.assertEquals("https://*", UrlWhiteListUtils.expandedWhiteListUrlString("https://*:*"));
+    }
+
+    @Test(expected = MalformedURLException.class)
+    public void expandIllegalWhitelistUrlString() throws MalformedURLException {
+        URL expURL = new URL(UrlWhiteListUtils.expandedWhiteListUrlString("https://subdomain.domain.com:1*"));
+    }
+
```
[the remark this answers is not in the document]
AndreasEhret: it would be also interesting if the MalformedURLException that is thrown could give the reason in its message, we can use the texts as defined in JIRA specs: Error: Invalid port, Error: Invalid host, etc. see OWS-383 (UI displayed column)

core/src/main/java/net/sourceforge/jnlp/util/UrlWhiteListUtils.java
```diff
@@ -0,0 +1,114 @@
...
+        } catch (Exception e) {
+            final int ind = expandedUrlStr.lastIndexOf(":");
+            if (e.getCause() instanceof NumberFormatException && expandedUrlStr.substring(ind+1, expandedUrlStr.length()).equals("*") ) {
+                expandedUrlStr =  expandedUrlStr.substring(0,ind) + "";
+            }
+        }
+        return expandedUrlStr;
+    }
+
+    public static String expandedWhiteListUrlString(final String wlUrlStr) {
```
AndreasEhret, line 109: could be package private

core/src/main/java/net/sourceforge/jnlp/util/UrlWhiteListUtils.java
```diff
@@ -0,0 +1,114 @@
...
+                        return UrlUtils.notNullUrlEquals(url, wlUrl);
+                    } else {
+                        final URL expandedWlUrl = new URL(UrlWhiteListUtils.expandedWhiteListUrlString(wlUrlStr));
+                        final boolean isUrlProtocolEqual = isUrlProtocolEqual(expandedWlUrl, url);
+                        final boolean isUrlHostWithWildcardEqual = isUrlHostWithWildcardEqual(expandedWlUrl, url);
+                        final boolean isUrlPortEqual = expandedWlUrl.getPort() != -1 ? expandedWlUrl.getPort() == url.getPort() : true;
+                        return isUrlProtocolEqual && isUrlHostWithWildcardEqual && isUrlPortEqual;
+                    }
+                } catch (Exception e) {
+                    LOG.debug("Bad white list url: " + wlUrlStr);
```
AndreasEhret, line 69: this is much more than a debug when the white list contains something invalid. Should be warning and log message should contain the information that this URL will be ignored.

hendrikebbers: +1

AndreasEhret: As we catch the exception here and just log we do not have the chance to show a warning decorator icon next to the whitelist in Settings UI with a tooltip "Your whitelist contains malformed entries" or similar to inform the user that something is not perfect with his configuration. Alternatively we could allow comments in white lists and whenever we found an invalid one, we insert it like
#Invalid Entry: <invalid entry>
that makes the problem visible to the user

sclassen:
merged this into branch `serverWhitelist` and opened new PR #685
I did this because the PR had conflicts with master after we merged #682

PR #675, opened 2020-07-15
download progress listener added

hendrikebbers:


hendrikebbers:
Will be used in https://github.com/karakun/OpenWebStart/pull/286

PR #679, opened 2020-07-30
Fixed EDT Classloading issues after EDT shutdown.

Gr33nbl00d:
The EventQueue.class stores the classloader in a final field classLoader. The EventQueue itself is stored static inside SunToolkit.class. Even if the active EDT thread classloader is modified to a JNLP classloader, classloading will fail once an AWTAutoShutdown occurs. (No edt activity for 1 second). If any new activity starts, the EDT Thread is recreated and initialized with the classloader stored in field classLoader in EventQueue class. This was not the JNLP classloader anymore. The result are ClassNotFound Exceptions. As there does not seem to be a "nice solution" because of the EventQueue implementation details the current solution temporary sets a DelegatingClassLoader to the thread which will initialize the eventqueue. Once initialized the original classloader is restored to the thread. During webstart loading the DelegatingClassLoader uses the original classloader once the jnlp classloader is active it gets injected into the DelegatingClassLoader. See:
https://github.com/karakun/OpenWebStart/issues/201

### DIFF
--- a/core/src/main/java/net/sourceforge/jnlp/runtime/AppContextFactory.java
+++ b/core/src/main/java/net/sourceforge/jnlp/runtime/AppContextFactory.java
@@ -8,13 +8,15 @@ public class AppContextFactory {
         //set temporary classloader for EventQueue initialization
         //already a call to AppContext.getAppContext(...) initializes the EventQueue.class
         ClassLoader originalLoader = Thread.currentThread().getContextClassLoader();
-        DelegatingClassLoader delegatingLoader = DelegatingClassLoader.getInstance();
-        delegatingLoader.setClassLoader(originalLoader);
+        try {
+            DelegatingClassLoader delegatingLoader = DelegatingClassLoader.getInstance();
+            delegatingLoader.setClassLoader(originalLoader);
 
-        Thread.currentThread().setContextClassLoader(delegatingLoader);
-
-        SunToolkit.createNewAppContext();
-        //restore original classloader
-        Thread.currentThread().setContextClassLoader(originalLoader);
+            Thread.currentThread().setContextClassLoader(delegatingLoader);
+            SunToolkit.createNewAppContext();
+        }finally {
+            //restore original classloader
+            Thread.currentThread().setContextClassLoader(originalLoader);
+        }
     }
 }
--- a/core/src/main/java/net/sourceforge/jnlp/runtime/AppContextFactory.java
+++ b/core/src/main/java/net/sourceforge/jnlp/runtime/AppContextFactory.java
@@ -3,10 +3,8 @@ package net.sourceforge.jnlp.runtime;
 import net.sourceforge.jnlp.runtime.classloader.DelegatingClassLoader;
 import sun.awt.SunToolkit;
 
-public class AppContextFactory
-{
-    public static void createNewAppContext()
-    {
+public class AppContextFactory {
+    public static void createNewAppContext() {
         //set temporary classloader for EventQueue initialization
         //already a call to AppContext.getAppContext(...) initializes the EventQueue.class
         ClassLoader originalLoader = Thread.currentThread().getContextClassLoader();
--- a/core/src/main/java/net/sourceforge/jnlp/runtime/AppContextFactory.java
+++ b/core/src/main/java/net/sourceforge/jnlp/runtime/AppContextFactory.java
@@ -1,0 +1,22 @@
+package net.sourceforge.jnlp.runtime;
+
+import net.sourceforge.jnlp.runtime.classloader.DelegatingClassLoader;
+import sun.awt.SunToolkit;
+
+public class AppContextFactory
+{
+    public static void createNewAppContext()
+    {
+        //set temporary classloader for EventQueue initialization
+        //already a call to AppContext.getAppContext(...) initializes the EventQueue.class
+        ClassLoader originalLoader = Thread.currentThread().getContextClassLoader();
+        DelegatingClassLoader delegatingLoader = DelegatingClassLoader.getInstance();
+        delegatingLoader.setClassLoader(originalLoader);
+
+        Thread.currentThread().setContextClassLoader(delegatingLoader);
+
+        SunToolkit.createNewAppContext();
+        //restore original classloader
+        Thread.currentThread().setContextClassLoader(originalLoader);
+    }
+}
--- a/core/src/main/java/net/sourceforge/jnlp/runtime/classloader/DelegatingClassLoader.java
+++ b/core/src/main/java/net/sourceforge/jnlp/runtime/classloader/DelegatingClassLoader.java
@@ -1,0 +1,89 @@
+package net.sourceforge.jnlp.runtime.classloader;
+
+import java.net.URL;
+
+public class DelegatingClassLoader extends ClassLoader
+{
+    public static final DelegatingClassLoader instance = new DelegatingClassLoader(Thread.currentThread().getContextClassLoader());
+    private ClassLoader classLoader;
+
+    public static DelegatingClassLoader getInstance()
+    {
+        return instance;
+    }
+
+    DelegatingClassLoader(ClassLoader loader)
+    {
+        super(loader);
+        this.classLoader = loader;
+    }
+
+    public void setClassLoader(ClassLoader loader)
+    {
+        this.classLoader = loader;
+    }
+
+    protected Class findClass(String name) throws ClassNotFoundException
+    {
+        return this.classLoader.loadClass(name);
+    }
+
+    protected URL findResource(String name)
+    {
+        return this.classLoader.getResource(name);
+    }
+
+    public int hashCode()
+    {
+
+        int result = 1;
+        result = 31 * result + (this.classLoader == null ? 0 : this.classLoader.hashCode());
+        result = 31 * result + (this.getParent() == null ? 0 : this.getParent().hashCode());
+        return result;
+    }
+
+    public boolean equals(Object obj)
+    {
+        if (this == obj)
+        {
+            return true;
+        }
+        else if (obj == null)
+        {
+            return false;
+        }
+        else if (this.getClass() != obj.getClass())
+        {
+            return false;
+        }
+        else
+        {
+            DelegatingClassLoader other = (DelegatingClassLoader) obj;
+            if (this.classLoader == null)
+            {
+                if (other.classLoader != null)
+                {
+                    return false;
+                }
+            }
+            else if (!this.classLoader.equals(other.classLoader))
+            {
+                return false;
+            }
+
+            if (this.getParent() == null)
+            {
+                if (other.getParent() != null)
+                {
+                    return false;
+                }
+            }
+            else if (!this.getParent().equals(other.getParent()))
+            {
+                return false;
+            }
+
+            return true;
+        }
+    }
+}
--- a/core/src/main/java/net/sourceforge/jnlp/runtime/classloader/DelegatingClassLoader.java
+++ b/core/src/main/java/net/sourceforge/jnlp/runtime/classloader/DelegatingClassLoader.java
@@ -19,8 +19,7 @@ public class DelegatingClassLoader extends ClassLoader {
     }
 
     public void setClassLoader(ClassLoader loader) {
-        Assert.requireNonNull(loader, "loader");
-        this.classLoader = loader;
+        this.classLoader = Assert.requireNonNull(loader, "loader");
     }
 
     protected Class findClass(String name) throws ClassNotFoundException {
--- a/core/src/main/java/net/sourceforge/jnlp/runtime/classloader/DelegatingClassLoader.java
+++ b/core/src/main/java/net/sourceforge/jnlp/runtime/classloader/DelegatingClassLoader.java
@@ -1,5 +1,7 @@
 package net.sourceforge.jnlp.runtime.classloader;
 
+import net.adoptopenjdk.icedteaweb.Assert;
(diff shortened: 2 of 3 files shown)


Gr33nbl00d:
Implementation of DelegatingClassLoader has same overriden methods than the original implementation of DelegationClassloader:
com.sun.xml.internal.ws.client.WSServiceDelegate.DelegatingLoader

Gr33nbl00d:
As i am not so deeply into classloading, i am asking myself it this implementation is correct from security perspective. Feel free to modify

hendrikebbers:
@Gr33nbl00d great to see this PR. Thank you very much. I will try to have a look at this next week.

core/src/main/java/net/sourceforge/jnlp/runtime/classloader/DelegatingClassLoader.java
```diff
@@ -0,0 +1,89 @@
...
+            return false;
+        }
+        else if (this.getClass() != obj.getClass())
+        {
+            return false;
+        }
+        else
+        {
+            DelegatingClassLoader other = (DelegatingClassLoader) obj;
+            if (this.classLoader == null)
```
sclassen, line 62: Please use `Objects.equals(a, b)` to compare two objects in a null safe way

Gr33nbl00d: Fixed

core/src/main/java/net/sourceforge/jnlp/runtime/classloader/DelegatingClassLoader.java
```diff
@@ -0,0 +1,73 @@
+package net.sourceforge.jnlp.runtime.classloader;
+
+import java.net.URL;
+import java.util.Objects;
+
+public class DelegatingClassLoader extends ClassLoader
+{
```
hendrikebbers, line 7: your format does not belong the the format that we use in the project . All '{' should be at the end of lines, for example.

core/src/main/java/net/sourceforge/jnlp/runtime/classloader/DelegatingClassLoader.java
```diff
@@ -0,0 +1,73 @@
...
+            return false;
+        }
+        else if (this.getClass() != obj.getClass())
+        {
+            return false;
+        }
+        else
+        {
+            DelegatingClassLoader other = (DelegatingClassLoader) obj;
+            if(Objects.equals(this.classLoader,other.classLoader)==false)
```
hendrikebbers, line 62:
```suggestion
            if(!Objects.equals(this.classLoader,other.classLoader))
```

core/src/main/java/net/sourceforge/jnlp/runtime/classloader/DelegatingClassLoader.java
```diff
@@ -0,0 +1,73 @@
...
+            {
+                return false;
+            }
+            if(Objects.equals(this.getParent(),other.getParent()) == false)
```
hendrikebbers, line 66:
```suggestion
            if(!Objects.equals(this.getParent(),other.getParent()))
```

core/src/main/java/net/sourceforge/jnlp/runtime/classloader/DelegatingClassLoader.java
Gr33nbl00d: I will change it manually, i would recommend to add a .editorconfig file to the repo so this problem will not occur anymore. https://editorconfig.org/

core/src/main/java/net/sourceforge/jnlp/runtime/classloader/DelegatingClassLoader.java
Gr33nbl00d: This is a personal thing :) I really dont like the ! because it is so tiny and quickly overlooked. But im perfectly ok if you want to have it that way

Gr33nbl00d: fixed

core/src/main/java/net/sourceforge/jnlp/runtime/classloader/DelegatingClassLoader.java
Gr33nbl00d: fixed

hendrikebbers: https://editorconfig.org/ looks like a good improvement. @sclassen what do you think? @Gr33nbl00d are you familiar with that and could provide a PR for it?

core/src/main/java/net/sourceforge/jnlp/runtime/classloader/DelegatingClassLoader.java
hendrikebbers: fine with it but in ITW / OWS we normally use the '!'. Thanks for fixing it.

core/src/main/java/net/sourceforge/jnlp/runtime/classloader/DelegatingClassLoader.java
Gr33nbl00d: Personally i use Intellij IDEA and there is an option to export the code style directly to .editorconfig. I guess that such an option is available for other IDEs too. So in best case you simply need to export your current config to .editorconfig and commit it to the repo.

core/src/main/java/net/sourceforge/jnlp/runtime/classloader/DelegatingClassLoader.java
```diff
@@ -0,0 +1,59 @@
...
+
+public class DelegatingClassLoader extends ClassLoader {
+    public static final DelegatingClassLoader instance = new DelegatingClassLoader(Thread.currentThread().getContextClassLoader());
+    private ClassLoader classLoader;
+
+    public static DelegatingClassLoader getInstance() {
+        return instance;
+    }
+
+    DelegatingClassLoader(ClassLoader loader) {
```
hendrikebbers, line 14:
```suggestion
    private DelegatingClassLoader(ClassLoader loader) {
```

core/src/main/java/net/sourceforge/jnlp/runtime/classloader/DelegatingClassLoader.java
```diff
@@ -0,0 +1,59 @@
...
+        super(loader);
+        this.classLoader = loader;
+    }
+
+    public void setClassLoader(ClassLoader loader) {
+        this.classLoader = loader;
```
hendrikebbers, line 20: Null Check is needed

core/src/main/java/net/sourceforge/jnlp/runtime/classloader/DelegatingClassLoader.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    public int hashCode() {
+        int result = 1;
+        result = 31 * result + (this.classLoader == null ? 0 : this.classLoader.hashCode());
+        result = 31 * result + (this.getParent() == null ? 0 : this.getParent().hashCode());
+        return result;
+    }
+
+    public boolean equals(Object obj) {
```
hendrikebbers, line 38: Since this is a singleton we simply check do `return this == obj` ;)

core/src/main/java/net/sourceforge/jnlp/runtime/AppContextFactory.java
```diff
@@ -0,0 +1,20 @@
...
+
+import net.sourceforge.jnlp.runtime.classloader.DelegatingClassLoader;
+import sun.awt.SunToolkit;
+
+public class AppContextFactory {
+    public static void createNewAppContext() {
+        //set temporary classloader for EventQueue initialization
+        //already a call to AppContext.getAppContext(...) initializes the EventQueue.class
+        ClassLoader originalLoader = Thread.currentThread().getContextClassLoader();
+        DelegatingClassLoader delegatingLoader = DelegatingClassLoader.getInstance();
```
hendrikebbers, line 11: all between line 10 and 18 should be in try { } block. Line 18 should be in finally {}

core/src/main/java/net/sourceforge/jnlp/runtime/classloader/DelegatingClassLoader.java
hendrikebbers: We use IntelliJ too. Can you give us link to a manual how to do that?

core/src/main/java/net/sourceforge/jnlp/runtime/classloader/DelegatingClassLoader.java
```diff
@@ -0,0 +1,59 @@
...
+public class DelegatingClassLoader extends ClassLoader {
+    public static final DelegatingClassLoader instance = new DelegatingClassLoader(Thread.currentThread().getContextClassLoader());
+    private ClassLoader classLoader;
+
+    public static DelegatingClassLoader getInstance() {
+        return instance;
+    }
+
+    DelegatingClassLoader(ClassLoader loader) {
+        super(loader);
```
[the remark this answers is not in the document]
Gr33nbl00d: fixed

core/src/main/java/net/sourceforge/jnlp/runtime/classloader/DelegatingClassLoader.java
Gr33nbl00d: fixed

hendrikebbers: more easy: `this.classLoader = Assert.requireNonNull(loader, "loader");`

core/src/main/java/net/sourceforge/jnlp/runtime/classloader/DelegatingClassLoader.java
```diff
@@ -0,0 +1,63 @@
...
+public class DelegatingClassLoader extends ClassLoader {
+    public static final DelegatingClassLoader instance = new DelegatingClassLoader(Thread.currentThread().getContextClassLoader());
+    private ClassLoader classLoader;
+
+    public static DelegatingClassLoader getInstance() {
+        return instance;
+    }
+
+    private DelegatingClassLoader(ClassLoader loader) {
+        super(loader);
```
hendrikebbers, line 17:
```suggestion
        super(Assert.requireNonNull(loader, "loader"));
```
More easy way to do it

core/src/main/java/net/sourceforge/jnlp/runtime/classloader/DelegatingClassLoader.java
Gr33nbl00d: You are right we could do it that way. However not sure if i would recommend it. Do you want me to change it?

hendrikebbers: Adding @sclassen to have an additional opinion.

core/src/main/java/net/sourceforge/jnlp/runtime/AppContextFactory.java
Gr33nbl00d: fixed

core/src/main/java/net/sourceforge/jnlp/runtime/classloader/DelegatingClassLoader.java
Gr33nbl00d: fixed

core/src/main/java/net/sourceforge/jnlp/runtime/classloader/DelegatingClassLoader.java
Gr33nbl00d: image

In the settings dialog a bit hidden maybe :)

hendrikebbers:
@FlorianRoettges & @Gr33nbl00d are you interested in a short Zoom call about OWS / ITW and this PR? You are German, right? If you are interested and have 5 min time for such call please send a short ping to openwebstart@karakun.com :)

core/src/main/java/net/sourceforge/jnlp/runtime/classloader/DelegatingClassLoader.java
hendrikebbers: https://github.com/karakun/OpenWebStart/issues/302

core/src/main/java/net/sourceforge/jnlp/runtime/classloader/DelegatingClassLoader.java
Gr33nbl00d: fixed

hendrikebbers:
@FlorianRoettges & @Gr33nbl00d thanks for taking care of all the mentioned points in the review. I like the current state. In general I think that the given "Workaround" (let's call it lime that) is a good way for ITW / OWS to handle the general problem. I will have a deeper look and additional tests with @sclassen before we merge this PR. Hope to have time for this next week. This would make it possible to have it in the August release of ITW & OWS. Have you seen my message above?

core/src/main/java/net/sourceforge/jnlp/runtime/classloader/DelegatingClassLoader.java
sclassen: Just leave the method as it is. Then it is safe for future refactoring

hendrikebbers:
@sclassen merge?

PR #680, opened 2020-07-31
add support for TSA Certificates

sclassen:


core/src/main/java/net/sourceforge/jnlp/tools/JarCertVerifier.java
```diff
@@ -303,10 +305,12 @@ private VerifyResult verifyJar(final String jarPath) {
     VerifyResult verifyJarEntryCerts(final String jarPath, final boolean jarHasManifest, final List<JarEntry> entries) {
         // Contains number of entries the cert with this CertPath has signed.
         final Map<CertPath, Integer> jarSignCount = new HashMap<>();
+        final Map<CertPath, CodeSigner> codeSigners = new HashMap<>();
         int numSignableEntriesInJar = 0;
 
         // Record current time just before checking the jar begins.
         final ZonedDateTime now = ZonedDateTime.now();
+        final ZonedDateTime shortlyAfterNow = now.plus(6, MONTHS);
```
hendrikebbers, line 313: 6 month are not "shortly after now" ;)

core/src/main/java/net/sourceforge/jnlp/tools/JarCertVerifier.java
```diff
@@ -342,15 +347,48 @@ VerifyResult verifyJarEntryCerts(final String jarPath, final boolean jarHasManif
...
-                        certInfo.setHasExpiredCert();
-                    } else if (now.plus(6, MONTHS).isAfter(notAfter)) {
-                        certInfo.setHasExpiringCert();
+
+                    final Optional<Timestamp> optionalSignatureTimestamp = Optional.ofNullable(codeSigners.get(certPath))
+                            .map(CodeSigner::getTimestamp);
+                    final X509Certificate tsaCertificate = (X509Certificate) optionalSignatureTimestamp
+                            .map(Timestamp::getSignerCertPath)
+                            .map(CertPath::getCertificates)
+                            .map(certs -> certs.get(0))
```
hendrikebbers, line 359: can this end in an ArrayOutOfBounds? Maybe a filter before that one makes sense.

core/src/main/java/net/sourceforge/jnlp/tools/JarCertVerifier.java
```diff
@@ -342,15 +347,48 @@ VerifyResult verifyJarEntryCerts(final String jarPath, final boolean jarHasManif
...
+                            .filter(c -> c instanceof X509Certificate)
+                            .orElse(null);
+
+                    if (tsaCertificate != null) {
+                        final ZonedDateTime tsaNotBefore = zonedDateTime(tsaCertificate.getNotBefore());
+                        final ZonedDateTime tsaNotAfter = zonedDateTime(tsaCertificate.getNotAfter());
+                        final ZonedDateTime signedAt = zonedDateTime(optionalSignatureTimestamp.get().getTimestamp());
```
hendrikebbers, line 366: We should use orElseThrow instead of get here. I understand that I can not happen now that the val is null but maybe in future based on refactoring. So maybe throw a specific exception + description here.

PR #682, opened 2020-08-03
Refactoring of resource download

hendrikebbers:


core/src/main/java/net/adoptopenjdk/icedteaweb/resources/ResourceHandler.java
```diff
@@ -34,8 +33,8 @@
 class ResourceHandler {
 
     private static final Logger LOG = LoggerFactory.getLogger(ResourceHandler.class);
-    private static final Executor localExecutor = new ThreadPoolExecutor(0, Runtime.getRuntime().availableProcessors() * 2,
-            10L, TimeUnit.SECONDS, new LinkedBlockingQueue<>());
+
+    private static final Executor localExecutor = Executors.newFixedThreadPool(Runtime.getRuntime().availableProcessors() * 4, new DaemonThreadFactory());
```
sclassen, line 37: Using a fixed thread pool for the download is not optimal.
Because after the initial download the threads are no longer used. Never the less the fixed thread pool will keep them idle until the end of the application

core/src/main/java/net/adoptopenjdk/icedteaweb/resources/ResourceTracker.java
```diff
@@ -388,20 +384,29 @@ private Resource getResource(URL location) {
...
      */
     private void wait(Resource... resources) throws InterruptedException {
-        // save futures in list to allow parallel start of all resources
-        final List<Future<Resource>> futures = Stream.of(resources)
-                .map(ResourceHandler::new)
-                .map(ResourceHandler::putIntoCache)
+        wait(resources, 100_000_000, TimeUnit.DAYS);
+    }
+
+    public static final String getSimpleName(Resource resource) {
```
sclassen, line 390: This method should be in the class `Resource`

core/src/main/java/net/adoptopenjdk/icedteaweb/resources/ResourceTracker.java
```diff
@@ -412,33 +417,35 @@ private void wait(Resource... resources) throws InterruptedException {
...
-
-            try {
-                future.get(waitTime, TimeUnit.NANOSECONDS);
-            } catch (TimeoutException e) {
-                return false;
-            } catch (ExecutionException ignored) {
-            }
-        }
-        return true;
+        return collectedDownloadFutures.stream()
```
sclassen, line 434: Not 100% sure but I believe the logic has changed in this block...
Maybe extract a method from the `map` block and give it a good name.
This may help to clarify if the logic is still the same

core/src/main/java/net/adoptopenjdk/icedteaweb/resources/ResourceHandler.java
hendrikebbers: I'm with you. This executor was just wrong and only executed one task in parallel. This is a quick fix to check that multiple downloads work in parallel. And yes, it works ;)

core/src/main/java/net/adoptopenjdk/icedteaweb/resources/ResourceTracker.java
hendrikebbers: ... or removed once I'm done ;) This just helped to create a lot of log.l Way to many log messages since we do not have trace logging

core/src/main/java/net/adoptopenjdk/icedteaweb/resources/ResourceTracker.java
hendrikebbers: will have a look

core/src/main/java/net/adoptopenjdk/icedteaweb/resources/ResourceHandler.java
```diff
@@ -34,43 +31,53 @@
...
+            } else {
+                LOG.debug("Download for {} has already been started.", resource.getSimpleName());
+                return future;
             }
-            resource.startProcessing(result);
         }
+    }
 
+    private Future<Resource> getDownloadStateAndStartUnstartedDownload(final Executor downloadExecutor) {
+        Assert.requireNonNull(downloadExecutor, "downloadExecutor");
```
sclassen, line 62: can you please move this check to the method `putIntoCache()`

core/src/main/java/net/adoptopenjdk/icedteaweb/resources/ResourceTracker.java
```diff
@@ -381,64 +364,61 @@ private Resource getResource(URL location) {
...
-                .map(ResourceHandler::putIntoCache)
-                .collect(Collectors.toList());
-
-        for (Future<Resource> future : futures) {
-            try {
-                future.get();
-            } catch (ExecutionException ignored) {
-            }
-        }
+    private Future<Resource> triggerDownloadFor(Resource resource, final Executor downloadExecutor) {
```
sclassen, line 367: please move this method one down (after `waitForCompletion(Resource... resources)`)

core/src/main/java/net/adoptopenjdk/icedteaweb/resources/ResourceTracker.java
```diff
@@ -381,64 +364,61 @@ private Resource getResource(URL location) {
...
-        final List<Future<Resource>> futures = Stream.of(resources)
-                .map(ResourceHandler::new)
-                .map(ResourceHandler::putIntoCache)
-                .collect(Collectors.toList());
+        final int configuredThreadCount = Integer.parseInt(JNLPRuntime.getConfiguration().getProperty(ConfigurationConstants.KEY_PARALLEL_RESOURCE_DOWNLOAD_COUNT));
+        final int threadCount = Math.min(configuredThreadCount, resources.length);
+        final ExecutorService downloadExecutor = Executors.newFixedThreadPool(threadCount, new DaemonThreadFactory());
+        try {
+            Arrays.asList(resources).stream()
+                    .map(r -> triggerDownloadFor(r, downloadExecutor))
```
sclassen, line 389: Do we need to collect **all** the futures in a list before we start calling `get()` on them?

core/src/main/java/net/adoptopenjdk/icedteaweb/resources/ResourceTracker.java
```diff
@@ -381,64 +364,61 @@ private Resource getResource(URL location) {
...
-        }
-        return true;
+    private int getPercentageDownloaded(URL[] allResources) {
+        return (int) Arrays.asList(allResources).stream()
+                .map(u -> getResource(u))
+                .mapToDouble(r -> {
+                    if (r.isComplete()) {
+                        return 100.0d;
+                    }
+                    if (r.isBeingProcessed()) {
```
sclassen, line 417: maybe check also for `r.getSize() > 0` to prevent division by zero

core/src/main/java/net/adoptopenjdk/icedteaweb/resources/ResourceTracker.java
hendrikebbers: What do you mean?

sclassen: The `triggerDownloadFor` returns a Future. This happens very fast (not blocking).
But on the next line the forEach will wait for the first Future to complete before the stream will process the next Rrsource.
Or at least that is how I think streams behave.

PR #683, opened 2020-08-03
add check for a trusted CA of the TSA certificate

sclassen:


core/src/main/java/net/sourceforge/jnlp/tools/JarCertVerifier.java
```diff
@@ -415,6 +415,24 @@ VerifyResult verifyJarEntryCerts(final String jarPath, final boolean jarHasManif
...
+            for (final Certificate c : certPath.getCertificates()) {
+                if (c instanceof X509Certificate) {
+                    final X509Certificate x509 = (X509Certificate) c;
+                    if (CertificateUtils.inKeyStores(x509, caKeyStores)) {
+                        return true;
+                    }
+                }
+            }
+        } catch (RuntimeException e) {
+            LOG.warn("Unable to read through CA cert store files.");
```
hendrikebbers, line 431: add e

hendrikebbers: and maybe add certPath in log?

core/src/main/java/net/sourceforge/jnlp/tools/JarCertVerifier.java
```diff
@@ -351,22 +351,22 @@ VerifyResult verifyJarEntryCerts(final String jarPath, final boolean jarHasManif
...
+                    final X509Certificate tsaCertificate = (X509Certificate) optionalTsa
                             .map(Timestamp::getSignerCertPath)
                             .map(CertPath::getCertificates)
                             .filter(certs -> !certs.isEmpty())
                             .map(certs -> certs.get(0))
                             .filter(c -> c instanceof X509Certificate)
                             .orElse(null);
 
-                    if (tsaCertificate != null) {
+                    if (tsaCertificate != null && isTrustedTsa(optionalTsa.map(Timestamp::getSignerCertPath).orElseThrow(() -> new RuntimeException("cert path is null even tough we have a tsa certificate")))) {
```
hendrikebbers, line 364: Maybe it makes sense to extract the optionalTsa before this if since we use it 2 times:

```
final Timestamp tsa = optionalTsa.orElseThrow(() -> new RuntimeException("...))
if(tsaCertificate != null && isTrustedTsa(tsa)) {
    // ...
    final ZonedDateTime signedAt = zonedDateTime(tsa.getTimestamp());
    //...
}
```

sclassen: I could not do the `orElseThrow` since it is absolutely OK for the optional to be empty in case that `tsaCertificate` is null

core/src/main/java/net/sourceforge/jnlp/tools/JarCertVerifier.java
sclassen: removed try catch since there is no method thrown anything, not even a RuntimeException

core/src/main/java/net/sourceforge/jnlp/tools/JarCertVerifier.java
hendrikebbers: maybe `.orElse(null)` ?

sclassen: then I think it would almost be simpler to not use the optional...

core/src/main/java/net/sourceforge/jnlp/tools/JarCertVerifier.java
```diff
@@ -415,6 +422,20 @@ VerifyResult verifyJarEntryCerts(final String jarPath, final boolean jarHasManif
         return result;
     }
 
+    private boolean isTrustedTsa(CertPath certPath) {
+        final List<KeyStore> caKeyStores = KeyStores.getCAKeyStores();
+        // Check entire cert path for a trusted CA
+        for (final Certificate c : certPath.getCertificates()) {
```
hendrikebbers, line 428: certPath can be null

core/src/main/java/net/sourceforge/jnlp/tools/JarCertVerifier.java
```diff
@@ -426,6 +447,9 @@ private VerifyResult verifySigners(final Map<CertPath, Integer> jarSignCount) {
     }
 
     private ZonedDateTime zonedDateTime(final Date date) {
+        if (date == null) {
```
hendrikebbers, line 450: this is not what I would assume for such method... maybe rename it

hendrikebbers:
@sclassen do you want to have this PR in 1.2 ?

core/src/main/java/net/sourceforge/jnlp/tools/JarCertVerifier.java
hendrikebbers: that it returns now if the param is null ;) Would not expect that. That this is not mandatory. Was just a comment

sclassen: I do not expect the date to be ever null.
But I am not 100% sure because the date comes from the X509 certificates and I do not know if there is a corner case where it would return null

hendrikebbers: let's stay with the current implementation

hendrikebbers:
@sclassen merge?

PR #686, opened 2020-08-06
server whitelist 2

sclassen:


core/src/main/java/net/sourceforge/jnlp/util/whitelist/WhitelistEntry.java
```diff
@@ -0,0 +1,99 @@
...
+
+/**
+ * ...
+ */
+public class WhitelistEntry {
+    private static final String WILDCARD = "*";
+    private static final String HOST_PART_REGEX = "\\.";
+
+    private final String rawWhitelistEntry;
+    private final URL effectiveWhitelistEntry;
```
hendrikebbers, line 17: does it make sense to have a `URL` type here? Even if it works now (and I do not get how `*` could be a valid port) it might break in future when we need to change something...

core/src/test/java/net/sourceforge/jnlp/util/whitelist/UrlWhiteListUtilsTest.java
```diff
@@ -14,35 +13,35 @@
...
+        Assert.assertEquals("https://subdomain.domain.com:9999", UrlWhiteListUtils.validateWhitelistUrl("https://subdomain.domain.com:9999").getEffectiveWhitelistEntry());
+        Assert.assertEquals("https://123.134.145.156:9999", UrlWhiteListUtils.validateWhitelistUrl("https://123.134.145.156:9999").getEffectiveWhitelistEntry());
+        Assert.assertEquals("http://123.134.145.156:8888", UrlWhiteListUtils.validateWhitelistUrl("http://123.134.145.156:8888").getEffectiveWhitelistEntry());
+
+        Assert.assertEquals("https://domain.com:443", UrlWhiteListUtils.validateWhitelistUrl("domain.com").getEffectiveWhitelistEntry());
+        Assert.assertEquals("https://*.domain.com:443", UrlWhiteListUtils.validateWhitelistUrl("*.domain.com").getEffectiveWhitelistEntry());
+        Assert.assertEquals("https://123.134.145.156:443", UrlWhiteListUtils.validateWhitelistUrl("123.134.145.156:443").getEffectiveWhitelistEntry());
+
+        Assert.assertEquals("http://subdomain.domain.com:80", UrlWhiteListUtils.validateWhitelistUrl("http://subdomain.domain.com").getEffectiveWhitelistEntry());
```
hendrikebbers, line 25: can you externalise an `assertEffectiveEntry` method? We do it several times and all lines are really long...

core/src/main/java/net/sourceforge/jnlp/util/whitelist/WhitelistEntry.java
```diff
@@ -0,0 +1,99 @@
...
+            return isUrlProtocolMatching(effectiveWhitelistEntry, url) && isUrlHostMatching(effectiveWhitelistEntry, url) && isUrlPortMatching(effectiveWhitelistEntry, url);
+        } else {
+            return false;
+        }
+    }
+    private static boolean isUrlProtocolMatching(URL url1, URL url2) {
+        return Objects.equals(url1.getProtocol(), url2.getProtocol());
+    }
+
+    private static boolean isUrlHostMatching(URL wlUrl, URL url) {
```
hendrikebbers, line 67: please us better names :)

core/src/main/java/net/sourceforge/jnlp/util/whitelist/WhitelistEntry.java
sclassen: No, but changing this would be a major refactor.
I love to do this but would prefer to do it in a separate PR

hendrikebbers: ok, than let's stay with URL for now

PR #692, opened 2020-08-30
show proxy error

janakmulani:


sclassen:
I am not sure if this is the right place to check the exception.
The URLProber is trying many different URLs some of which will be not valid.

Maybe it is better to check the exception in BaseResourceDownloader.
And even then only if none of the URLs was able to download the resource

janakmulani:
@sclassen 
> I am not sure if this is the right place to check the exception.
> The URLProber is trying many different URLs some of which will be not valid.
> 
> Maybe it is better to check the exception in BaseResourceDownloader.
> And even then only if none of the URLs was able to download the resource

If there is a proxy error no URL would succeed. At present Proxy errors are caught in HttpURLConnection and handled in WebStartProxySelector.connectionFailed. Since consumeAndCloseConnectionSilently is not throwing an exception. UrlProber seems to be the only place where we can flag a Proxy error. Nevertheless, let's discuss.

sclassen:
Can you test if the IO exception could also be catched in the `BaseResourceDownloader.futureToOptional()`.
If this is possible than we could raise a flag in the BaseResourceDownloader and if none of the URLs in the `download()` method succeeds we could throw pop up the dialog on line 70 in BaseResourceDownloader

janakmulani:
@sclassen  Please review.

sclassen:
I slightly restructured the code such that it has less diff to master.
Works great on my machine. Thanks

PR #704, opened 2020-09-22
Jvmargs issues

janakmulani:


core/src/main/java/net/adoptopenjdk/icedteaweb/jvm/JvmUtils.java
```diff
@@ -41,29 +39,24 @@
      * @throws IllegalArgumentException if the VM arguments are invalid or dangerous
      */
     public static void checkVMArgs(final String vmArgs) throws IllegalArgumentException {
-        final boolean isJava9orGreater = JAVA_9_OR_GREATER.contains(JVM_VERSION);
-        checkVMArgs(vmArgs, isJava9orGreater);
-    }
-
-    // visible for testing
-    static void checkVMArgs(final String vmArgs, final boolean isJava9orGreater) {
```
hendrikebbers, line 49: I know that the method already was designed like this but maybe it makes sense to provide a return value instead of just throwing an exception.

What I would like:

`static Collection<VmArgumentViolation> checkVMArgs(final String vmArgs)`

The returned collection is never null and contains violations for each arg that is not allowed. `VmArgumentViolation` contains the arg name (and maybe the index).

core/src/main/java/net/adoptopenjdk/icedteaweb/jvm/JvmUtils.java
```diff
@@ -41,29 +39,24 @@
...
-        if (isBlank(vmArgs)) {
+       if (isBlank(vmArgs)) {
             return;
         }
 
         final String[] arguments = vmArgs.trim().split("\\s+");
```
hendrikebbers, line 46: Can we create a constant for this string? <- "\\s+"

core/src/main/java/net/adoptopenjdk/icedteaweb/jvm/JvmUtils.java
```diff
@@ -110,6 +103,17 @@ public static boolean isValidSecureProperty(final String argument) {
         return VALID_SECURE_PROPERTIES.contains(argument);
     }
 
+    // Arguments that are specified in deployment.jvm.arguments.whitelist in deployment.config
+    private static boolean isValidConfigArgument(final String argument) {
+        final String[] argParts = argument.split("=");
```
hendrikebbers, line 108: please provide a constant <- "="

janakmulani:
There seems to be a problem with a test which passes locally. I will have a look.

core/src/main/java/net/adoptopenjdk/icedteaweb/jvm/JvmUtils.java
```diff
@@ -110,6 +103,17 @@ public static boolean isValidSecureProperty(final String argument) {
...
+        for (String configArgument : CONFIG_JVM_ARGS) {
```
hendrikebbers, line 109: return CONFIG_JVM_ARGS.stream()
           .anyMatch(v -> Objects.equals(v, argParts[0]));

hendrikebbers (commented):
Minimal change requests

sclassen:
PR looks good.
But it is really hard to read because it contains 3 or 4 different changes.
Next time please make separate PR for each change.

core/src/main/java/net/adoptopenjdk/icedteaweb/jvm/JvmUtils.java
janakmulani: it is called from Parser. didn't want to touch it. besides the logic is that if one arg is bad all args are ignored.

janakmulani:
> 
> 
> PR looks good.
> But it is really hard to read because it contains 3 or 4 different changes.
> Next time please make separate PR for each change.
I thought making separate commits will help but apparently not.

@sclassen Should this property be supported in response.varfile? I guess not.

core/src/main/java/net/adoptopenjdk/icedteaweb/jvm/JvmUtils.java
```diff
@@ -276,7 +287,10 @@ public static boolean isValidSecureProperty(final String argument) {
                 "sun.java2d.win.uiScaleX",
                 "sun.java2d.win.uiScaleY",
                 "sun.java2d.uiScale",
-                "prism.allowhidpi" // for JavaFX
+                "prism.allowhidpi", // for JavaFX
+                "sun.net.client.defaultConnectTimeout",
+                "sun.net.client.defaultReadTimeout",
+                "java.locale.providers"
```
sclassen, line 293: Can you remove the `java.locale.providers`.
See OWS-465 for the security relevant restrictions.

Best to put this in a separate PR

core/src/test/java/net/adoptopenjdk/icedteaweb/jvm/JvmUtilsTest.java
```diff
@@ -242,4 +218,35 @@ public void testMoreThanOneBlankBtwnArgs() {
...
+        final String java_vm_args = "-XX:+UseParNewGC -XX:+CMSParallelRemarkEnabled -Dsun.net.client.defaultConnectTimeout=5000 -Dsun.net.client.defaultReadTimeout=500 -Djava.locale.providers==COMPAT,SPI";
+        try {
+            JvmUtils.checkVMArgs(java_vm_args);
+        } catch (IllegalArgumentException ile) {
+            fail(ile.getMessage());
+        }
+    }
+
+    @Before
+    public void setConfigJvmArgs() {
```
sclassen, line 233: Please move this method to the top of the class.
Otherwise this will not be noticed by other devs.

sclassen (changes requested):

puneet1725:
I am using Java version 17 and using java-vm-args="--add-opens=java.desktop/sun.font=ALL-UNNAMED in openwebstart.
But it doesn't work. is this issue being solved here?

Issue #700, opened 2020-09-17
Add Travis Build


hendrikebbers:
**🆕🐥🐶 First Timers Only**
This issue is reserved for people who never contributed to Open Source before. We know that the process of creating a pull request is the biggest barrier for new contributors. This issue is for you 💝

**👾 Description of the issue**
IcedTea-Web should have a [Travis](https://travis-ci.org) build. Information how a Java project can be build by Travis can be found [here](https://docs.travis-ci.com/user/languages/java/). IcedTea-Web is based on [Maven](https://maven.apache.org) and a `mvn clean verify` should be executed by the Travis build. Today we do not have a maven wrapper but that can be added when solving this issue (optional). IcedTea-Web should be build on the last Java 8 release of [AdoptOpenJDK](https://adoptopenjdk.net). Providing an additional build that is based on Java 11 would be a big benefit (optional).

**📋 Step by Step**
To solve this issue and contribute a fix to IcedTea-Web you should check the following step-by-step list:

* Claim this issue: Comment below.
* fork the repository in github by simply clicking the 'fork' button.
* check out the forked repository
* create a feature branch for the issue. We do not have any naming definition for branches.
* Commit your changes.
* Start a Pull Request.
* Done 👍 Ask in comments for a review :)
* If the reviewer find some missing peaces or a problem he will start a discussion with you and describe the next steps how the problem can be solved.
* you did it! We will merge the fix in the master of the IcedTeaWeb project.
* Thanks, thanks, thanks for being part as an open source contributor for IcedTeaWeb

**🎉 Contribute to Hacktoberfest**
Solve this issue as part of the [Hacktoberfest](https://hacktoberfest.digitalocean.com) event and get a change to receive cool goodies like a T-Shirt. 🎽

**🤔❓ Questions**
If you have any questions just ask us directly in this issue by adding a comment. You can join our community chat at [Slack](https://adoptopenjdk.net/slack.html). Next to this you can find a general manual about open source contributions here.

karianna:
Commenting purely on the template design here (if we were to adopt it across all repos).  I do find the heavy icon usage distracting.  Perhaps we could tone down some of that.

The bulleted list could/should be numeric?

Overall I love the concept

hendrikebbers:
Added the template as a GIST for future discussion here: https://gist.github.com/hendrikebbers/8e4dec9ddea5e2a420080d1314af025f

[...]

PR #708, opened 2020-10-01
Travis integration

bmsbharathi:
Added AdoptJDK8 installation in Travis-CI integration.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -3,6 +3,7 @@ Please keep this list in alphabetical order.
 
 Lillian Angel <langel@redhat.com>
 Andrew Azores <aazores@redhat.com>
+Bharathi BMS <bmsbharathi@gmail.com>
 Deepak Bhole <dbhole@redhat.com>
 Laurent Bourgès <bourges.laurent@gmail.com>
 Adam Buchta <adbuch7@gmail.com>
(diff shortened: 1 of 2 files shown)


sclassen:
The updating and installing of the dependencies adds about 1 Minute to the build time.
Maybe using sdkman is faster.
We would need to check this in a separate branch

bmsbharathi:
@sclassen hey! 
Just changed the JDK installation through sdkman.

AUTHORS
```diff
@@ -3,6 +3,7 @@ Please keep this list in alphabetical order.
 
 Lillian Angel <langel@redhat.com>
 Andrew Azores <aazores@redhat.com>
+Bharathi BMS <bmsbharathi@gmail.com>
```
hendrikebbers, line 6: While it is totally ok to add your name here the file is not really in sync anymore and I assume we will remove it in future :)

hendrikebbers:
link to https://github.com/AdoptOpenJDK/IcedTea-Web/issues/700

sclassen:
greate !!
SDK man brings a significant performance improvement over the deb packages.
Could you as a last step add the SDK man installation directory to the cache. Such that the JVM does not need to be re-downloaded from the internet.

sclassen:
@bmsbharathi just want to quickly check back if you need any assistance with adding the SDK dir to the cache?

PR #711, opened 2020-10-07
fix: JARs downloaded with URI-escaped characters don't get loaded

### LINKED ISSUE #573: JARs downloaded with URI-escaped characters don't get loaded
If a `jnlp` file is loaded that contains a JAR like `https://example.com/thing.jar?foo=bar%40baz`, the file gets stored in `.cache/icedtea-web/cache/N/https/example.com/thing.jar.foo_bar%40baz`, but but when the URI classloader tries to load the JAR, it doesn't load it.

I'm not certain on the exact details, but in 1.8, adding `%` to the `FileUtils.INVALID_PATH` list seems to fix it? This is 1.8.3 with the OpenJDK JRE on Linux (NixOS)

SchwingSK:
Closes #573


sclassen:
Thank you for this PR.
We will have a look and come back to you soon.

PR #723, opened 2020-11-05
Reduced number of threads spawned for downloading of resources. Speedup is aprox 200-300%

judovana:
For detailed results see: https://github.com/judovana/GenerateUnlimitedNumberOfJars
Lmited number of visible download bars
Added proeprties for backward compatibility and fortuning above
Originally by akaschen@redhat.com

### DIFF
--- a/netx/net/sourceforge/jnlp/Launcher.java
+++ b/netx/net/sourceforge/jnlp/Launcher.java
@@ -552,7 +552,7 @@ public class Launcher {
             }
 
             OutputController.getLogger().log(OutputController.Level.ERROR_ALL, "Starting application [" + mainName + "] ...");
-            
+
             Class<?> mainClass = app.getClassLoader().loadClass(mainName);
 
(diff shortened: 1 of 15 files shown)


judovana:
When I look into the results, then it is obvious that ITW  2.0 have already change like this incorporated.
Thus my main reason for this review, is to try to align with 2.0 as much as possible.

sclassen:
@judovana will merge this as it looks good.
Do you see something we should incorporate into the 2.x branch to gain some speedup there?

judovana:
ITW.2.0 have already this patch inside - in sme or other form. See the results: https://github.com/judovana/GenerateUnlimitedNumberOfJars

Only idea was to maybe disuc the swithces. But most of them is nto necessary.


Is itw 2.0 also reducing the visible donwload bars in huge apps?

judovana:
Will try to move it also to 1.7 , as it is used form time to time on some obscurer  systems.

PR #727, opened 2020-12-02
correct name of deployment config file

janakmulani:


sclassen:
Thank you Janak, this was a more than stupid mistake from my side :slightly_frowning_face:

PR #729, opened 2020-12-15
Better HTTPS with Client Authentication & SmartCard support on Windows

DavidTavoularis:
a. Better HTTPS with Client authentication :
 - In order to avoid too many PIN/Password requests (HTTPS hanshakes), downloads first the initial jar (mono-thread) then all remaining ones (multi-thread)
 - In case of initial jar download failure (Client Certificate Selection cancel, incorrect PIN/Password, ...), fail fast and do not try to download remaining jars
 - No resource prefetch, otherwise deployment.cache.parallelDownloadCount is not enforced

 b. New Security Dialog UI for Client Certificate Selection :
 - Ability to display details on each certificate
 - Shown in TaskBar
 - Cancellable
 - Selection using double-click, Enter key or OK button
 - Internationalization

 c. SmartCard support on Windows
 - Introduction of a Merged Key Manager for user, system and browser key stores
 - Client alias selection using a preference algorithm :
 a) Get all non-expired aliases with extension ClientCert (1.3.6.1.5.5.7.3.2) or ANY (2.5.29.37.0)
 if only one, it is selected, if more than one, a UI helps to select one of them.
 b) Otherwise, get all expired aliases with extension ClientCert (1.3.6.1.5.5.7.3.2) or ANY (2.5.29.37.0)
 if only one, it is selected, if more than one, a UI helps to select one of them.
 c) Otherwise, get all remaining aliases
 if only one, it is selected, if more than one, a UI helps to select one of them.
 - UI will display the suffix " (from user keystore)" or " (from system keystore)" or " (from browser keystore)"
 - If user cancels the Client Certificate UI, it is remembered for next chooseClientAlias call
 - Client alias selection is cached per remote host

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -39,6 +39,7 @@ Ville Skyttä <ville.skytta@iki.fi>
 Fridrich Strba <fridrich.strba@suse.com>
 Andrew Su <asu@redhat.com>
 Joshua Sumali <jsumali@redhat.com>
+David Tavoularis <david.tavoularis@mycom-osi.com>
 Joel Tesdall <jtesdall@mapcon.com>
 Michal Vala <mvala@redhat.com>
 Jiri Vanek <jvanek@redhat.com>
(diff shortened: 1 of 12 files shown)


sclassen:
Merging this into a feature branch in order to experiment with the code.

PR #739, opened 2021-01-19
Flag to disable restricted files

sclassen:


sclassen:
I had to extract the extract the method `createRestrictedFile` into a new class and move this new class to another sub-project.
Otherwise I would not be able to access the configuration because it is in a different module.

janakmulani:
> 
> 
> I had to extract the extract the method `createRestrictedFile` into a new class and move this new class to another sub-project.
> Otherwise I would not be able to access the configuration because it is in a different module.

Looks OK.

AndreasEhret (approved):

PR #745, opened 2021-02-05
code cleanup and do not treat windows special

sclassen:


common/src/main/java/net/adoptopenjdk/icedteaweb/lockingfile/LockableFile.java
```diff
@@ -88,12 +91,17 @@ public static LockableFile getInstance(final File file) {
...
         this.file = file;
         try {
-            //just try to create
-            this.file.createNewFile();
+            if (! file.exists()) {
+                if (! file.createNewFile()) {
+                    throw new IOException("could not create file " + file);
+                }
+            } else if (! file.isFile()) {
+                logger.error("lockable file {} is not a file but something else (maybe a directory)", file);
```
hendrikebbers, line 99: why not throwing an exception?

common/src/main/java/net/adoptopenjdk/icedteaweb/lockingfile/LockableFile.java
```diff
@@ -88,12 +91,17 @@ public static LockableFile getInstance(final File file) {
...
+            }
         } catch (final Exception ex) {
-            //intentionally silent
+            logger.error("Exception while creating lockable file", ex);
```
hendrikebbers, line 102: In general I ask myself if throwing exceptions would not make more sense. But this would maybe change the runtime behavior...

core/src/main/java/net/sourceforge/jnlp/util/PropertiesFile.java
```diff
@@ -216,16 +215,19 @@ public boolean load() {
...
         try (final FileOutputStream s = new FileOutputStream(file)) {
-            file.getParentFile().mkdirs();
             delegate.store(s, header);
             // fsync()
             s.getChannel().force(true);
             lastStore = file.lastModified();
             dirty = false;
         } catch (final IOException ex) {
-            LOG.error(IcedTeaWebConstants.DEFAULT_ERROR_MESSAGE, ex);
+            LOG.error("Failed to store", ex);
```
hendrikebbers, line 230: In general I ask myself if throwing exceptions would not make more sense. But this would maybe change the runtime behavior...

sclassen:
These changes break ITW on Windows completely as the JVM on Windows will not release the file lock in a timely manner.
Thus blocking the file for other processes for an indefinite time and thereby preventing multiple JNLPs to be run a the same time...
see: https://bugs.java.com/bugdatabase/view_bug.do?bug_id=4715154

common/src/main/java/net/adoptopenjdk/icedteaweb/lockingfile/LockableFile.java
sclassen: Yes this is the main reason, try to alter the current behavior as little as possible

PR #748, opened 2021-02-11
actions: add linux build for 1.8 branch

gdams:


hendrikebbers:
This should be used for the 1.8 branch. For master we have https://github.com/AdoptOpenJDK/IcedTea-Web/blob/master/.github/workflows/maven.yml @sclassen @gdams Does anybody know if we can have specific / different actions for PRs based on the target branch?

gdams:
@hendrikebbers yes you can, in the config I've told it to only
Run on PR's targeting the 1.8 branch

PR #749, opened 2021-02-12
implement tryLock and wait for free lock

sclassen:


AndreasEhret (approved):
Looks good and tests are green

PR #751, opened 2021-02-13
the path to user's desktop folder is read from windows registry

vnashkev:
Currently the path to user's desktop folder is defined as 
```     
private String getDesktopLnkPath() {
        return System.getenv("userprofile") + "/Desktop/" + getShortcutFileName();
}
```
This is not always true.
It seems to be more reliable to get desktop path from windows registry
```     
HKEY_CURRENT_USER\SOFTWARE\Microsoft\Windows\CurrentVersion\Explorer\Shell Folders\Desktop
```
Current definition may be used as a fallback if registry read fails or is impossible for some reason.

This pull request will probably fix https://github.com/karakun/OpenWebStart/issues/243 issue.

### DIFF
--- a/core/src/main/java/net/sourceforge/jnlp/util/WindowsDesktopEntry.java
+++ b/core/src/main/java/net/sourceforge/jnlp/util/WindowsDesktopEntry.java
@@ -37,6 +37,8 @@ import java.nio.file.Files;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
 
+import javax.swing.filechooser.FileSystemView;
+
 import static java.nio.charset.StandardCharsets.UTF_8;
 import static net.adoptopenjdk.icedteaweb.IcedTeaWebConstants.DOUBLE_QUOTE;
 import static net.sourceforge.jnlp.util.WindowsShortcutManager.getWindowsShortcutsFile;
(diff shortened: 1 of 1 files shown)


core/src/main/java/net/sourceforge/jnlp/util/WindowsDesktopEntry.java
```diff
@@ -204,4 +207,36 @@ public File getLinuxMenuIconFile() {
...
+							final String desktopPath = output.substring(typeIndex + typeName.length());
+							if (desktopPath != null) {
+								return desktopPath.trim();
+							}
+						}
+					}
+				}
+			}
+		} catch (final Exception e) {
+			LOG.error("Could not retrieve path to client desktop folder from registry", e);
```
sclassen, line 238: Can you make this log just a warning

core/src/main/java/net/sourceforge/jnlp/util/WindowsDesktopEntry.java
```diff
@@ -204,4 +207,36 @@ public File getLinuxMenuIconFile() {
...
+		}
+		return System.getenv("userprofile") + "/Desktop";
```
sclassen, line 240: Instead of using this trivial implementation which is known to fail could you implement the fallback using the method proposed in the issue https://github.com/karakun/OpenWebStart/issues/243

```
FileSystemView.getFileSystemView().getHomeDirectory().getAbsolutePath()
```

sclassen (changes requested):
Hi, and thank you for your contribution to icedtea-web.

I have two minor change requests. Otherwise the code looks great.

sclassen:
Thanks again for your contribution and have a wonderful day  :smiley:

vnashkev:
With great plesure.
Thank you.

PR #752, opened 2021-02-15
Allow to specify require 32 bit JVM in JNLP file

AndreasEhret:
Provide an attribute to be able to implement logic to use a 32 bit JVM on a 64 bit Windows machine

### DIFF
--- a/core/src/main/java/net/sourceforge/jnlp/Parser.java
+++ b/core/src/main/java/net/sourceforge/jnlp/Parser.java
@@ -502,11 +502,12 @@ public final class Parser {
             vmArgs = null;
         }
         final String vendor = getAttribute(node, JREDesc.VENDOR_ATTRIBUTE, null);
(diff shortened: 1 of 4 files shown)


core/src/main/java/net/sourceforge/jnlp/Parser.java
```diff
@@ -502,11 +502,12 @@ private JREDesc getJRE(final XmlNode node) throws ParseException {
             vmArgs = null;
         }
         final String vendor = getAttribute(node, JREDesc.VENDOR_ATTRIBUTE, null);
+        final boolean require32Bit = "true".equals(getAttribute(node, JREDesc.REQUIRE_32_BIT_ATTRIBUTE, "false"));
```
sclassen, line 505: Can you use `equalsIgnoreCase` or `Boolean.parseBoolean()`

sclassen (changes requested):

core/src/main/java/net/sourceforge/jnlp/Parser.java
AndreasEhret: done

sclassen (approved):

PR #759, opened 2021-02-17
add .gitattributes file

### LINKED ISSUE #757: bash files should be forced to LF line endings
When the project is cloned in Windows, all *.sh files have CRLF line endings, which make building of the launchers under cygwin fail (without any helpful message). It can be easily done with a .gitattributes file

SchwingSK:
Fixes #757


sclassen:
Thank you for this fix

PR #763, opened 2021-02-18
add windows signing

gdams:


gdams:
Verified that the release is codesigned: 

Screenshot 2021-02-18 at 14 52 46

douph1 (changes requested):
If intended to produce release, please consider https download instead of http 
http://ftp.mozilla.org/pub/mozilla.org/js/rhino1_6R7.zip

For all deps, please considere adding signing check or checksum check of runtime downloaded deps ..
rhino1_6R7.zip
AzureSignTool.zip

douph1:
and same suggestion for https://github.com/AdoptOpenJDK/IcedTea-Web/pull/761/
use https instead of http for downloading tagsoup-1.2.jar.zip and rhino1_6R7.zip
and consider checking checksum for setup-x86_64.exe, tagsoup-1.2.jar.zip, mslinks-1.0.5.jar, wixgen.jar, rhino1_6R7.zip

gdams:
@douph1 updated, I've added shasum checks to all downloads (except cygwin - it's not a fixed link so changes all the time)

Also updated to use https where possible

douph1:
You can probably check the signature for Cygwin as the sig if provided and the key too : 1A698DE9E2E56300
https://cygwin.com/install.html

douph1 (changes requested):
This deps is HTTP not HTTPS http://www.java2s.com/Code/JarDownload/tagsoup/tagsoup-1.2.jar.zip
and this is not what it pretend to be : the META-INF inside the jar say it is Version: 1.1.3
not 1.2 !!
It scares me.

I promptly recommend to download from other well known place like Maven Central : https://repo1.maven.org/maven2/org/ccil/cowan/tagsoup/tagsoup/1.2.1/
where the 1.2.1 seems to be a clone of original 1.2.1 ( with META-INF version who match)
and with signature and checksum :)
produce by https://github.com/jukka/tagsoup

gdams:
@douph1 updated tagsoup URL and added gpg verification of cygwin.

.github/workflows/build.yml
```diff
@@ -155,7 +158,10 @@ jobs:
       - name: Install Cygwin
         run: |
           Invoke-WebRequest -UseBasicParsing 'https://cygwin.com/setup-x86_64.exe' -OutFile 'C:\temp\cygwin.exe'
-          Start-Process -Wait -FilePath 'C:\temp\cygwin.exe' -ArgumentList '--packages autoconf,automake,cpio,curl,gcc,git,grep,libtool,make,mingw64-x86_64-gcc-core,perl,pkg-config,unzip,wget,zip --quiet-mode --download --local-install --delete-orphans --site https://mirrors.kernel.org/sourceware/cygwin/ --local-package-dir C:\cygwin_packages --root C:\cygwin64'
+          Invoke-WebRequest -UseBasicParsing 'https://cygwin.com/setup-x86_64.exe.sig' -OutFile 'C:\temp\cygwin.exe.sig'
+          gpg --keyserver keys.gnupg.net --recv-key 676041BA
```
douph1, line 162: Why choosing the old key ? 
>During the transition period, signatures are made using both old (676041BA) and new (1A698DE9E2E56300) public >keys here.

douph1 (approved):

.github/workflows/build.yml
gdams: hmm I tried with the new key and it failed, looks like the old key failed too

gdams: @douph1 any idea why the gpg steps are failing?

gdams: Looks like `keys.gnupg.net` might be down?

```
➜  IcedTea-Web git:(1.8) gpg --recv-key 1A698DE9E2E56300
gpg: keyserver receive failed: No data
➜  IcedTea-Web git:(1.8) ping keys.gnupg.net
ping: cannot resolve keys.gnupg.net: Unknown host
```

gdams: ahh it needs both...

PR #775, opened 2021-02-24
Do not check permissions for JNLP files without jars

### LINKED ISSUE #774: Jar permissions should not be checked when the JNLP resource contains no jars
This is a corner case, but it affects our application:
our software is deployed on our clients' servers, and we have a plugin system where our clients can add their own jars to add functionalities. To this end we have our JNLP that includes another JNLP (so that the client can sign their jars with their own certificate, and facilitate maintenance):

            <extension href="jarBaseUrl/external.jnlp" />
The client can put whatever jars he wants in this second JNLP file.
However, by default 'external.jnlp' contains no jars. So for our clients that do not provide their own jars, they get a dialog warning them that there's a problem with the permissions, when there should be no such warning (as is the case with Oracle's JavaWebStart).

SchwingSK:
Fixes #774

### DIFF
--- a/core/src/main/java/net/adoptopenjdk/icedteaweb/manifest/ManifestAttributesChecker.java
+++ b/core/src/main/java/net/adoptopenjdk/icedteaweb/manifest/ManifestAttributesChecker.java
@@ -116,7 +116,7 @@ public class ManifestAttributesChecker {
 
             if (attributesCheck.contains(MANIFEST_ATTRIBUTES_CHECK.PERMISSIONS) ||
                     attributesCheck.contains(MANIFEST_ATTRIBUTES_CHECK.ALL)) {
-                checkPermissionsAttribute();
+            	checkPermissionsAttribute();
             } else {
                 LOG.warn("check on {} skipped because property of deployment.manifest.attributes.check was not set to ALL or includes {} in the combination of options", "Permissions", "PERMISSIONS");
             }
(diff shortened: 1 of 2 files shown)


core/src/main/java/net/adoptopenjdk/icedteaweb/manifest/ManifestAttributesChecker.java
```diff
@@ -116,7 +116,11 @@ public void checkAll() throws LaunchException {
 
             if (attributesCheck.contains(MANIFEST_ATTRIBUTES_CHECK.PERMISSIONS) ||
                     attributesCheck.contains(MANIFEST_ATTRIBUTES_CHECK.ALL)) {
-                checkPermissionsAttribute();
+                if (file.getResources().getJARs().length > 0) {
```
sclassen, line 119: Could you please move this check into the method `checkPermissionsAttribute()`

sclassen (changes requested):
Hello, welcome and thank you for your contribution to icedtea-web.

I have only a minor change request. It would be great if you could update your PR.

core/src/main/java/net/adoptopenjdk/icedteaweb/manifest/ManifestAttributesChecker.java
SchwingSK: No problem, done.

Issue #782, opened 2021-03-12
Exception about duplicate class definition during start


VolvoDev:
Error occurs when we start IceadTea-Web with Volvo RTDMS project application. 
Exception:

_Exception in thread "Thread-8" java.lang.LinkageError: loader (instance of  net/sourceforge/jnlp/runtime/JNLPClassLoader): attempted  duplicate class definition for name: "org/apache/commons/logging/impl/LogFactoryImpl"_ 

Full stack trace is available in attachement:
[stacktrace.txt](https://github.com/AdoptOpenJDK/IcedTea-Web/files/6129871/stacktrace.txt).

Possible cause: Not properly synchronized net.sourceforge.jnlp.runtime.classloader.JNLPClassLoader.loadClass(final String name)

PR #783, opened 2021-03-12
Fixing issue 782 -> Exception about duplicate class definition during…

VolvoDev:
Hi,

This is a fix for https://github.com/AdoptOpenJDK/IcedTea-Web/issues/782
It was already tested inside Volvo project and working fine.

Just adding thread synchronization on class loader, to prevent duplicate class loading issue.
Thread synchronization is made in same manner as in parent class java.lang.ClassLoader#loadClass(java.lang.String, boolean), so the loc it's per loaded class, and does not affect efficency.

Patryk


sclassen:
Have you tested that this change will fix your class loader issue?
It could also be that the setup of your JNLP is causing this and there is a fix without adding yet another synchronized block

VolvoDev:
Hi @sclassen 
We were testing this fix extensively in our project, it worked fine. 
I did not run unit test.

JNLP file is totally correct as it's working from many years with Oracle Java Web Start.

So this code was synchronized on purpose, moreover thread synchronization is per loaded class, so it will not affect performance, and simple fix issue that we have.

We have a specific (quite big) Java client project with something like +8300 classes + libraries to be loaded, so this is a good tester :)

Let me know in case if you have more questions.

Thanks!
Patryk Kałużny

PR #803, opened 2021-05-17
Broken cache files during download

sclassen:


janakmulani:
@sclassen I am looking at this.

core/src/main/java/net/adoptopenjdk/icedteaweb/resources/downloader/BaseResourceDownloader.java
```diff
@@ -149,7 +149,15 @@ private long tryDownloading(final DownloadDetails downloadDetails) throws IOExce
             final DownloadInfo downloadInfo = new DownloadInfo(resourceHref, version, downloadDetails.lastModified);
             final File cacheFile = Cache.addToCache(downloadInfo, unpackedContent);
             resource.setLocalFile(cacheFile);
-            return countingInputStream.numBytesRead();
+
+            final long expectedBytes = downloadDetails.totalSize;
+            final long actualBytes = countingInputStream.numBytesRead();
+
+            if (expectedBytes > 0 && expectedBytes > actualBytes) {
+                throw new IOException(String.format("Did read %d bytes from server but expected %d", actualBytes, expectedBytes));
```
janakmulani, line 157: Can we not store all such files in a list and delete them all from cache? We know that these files will likely cause issues like ClassNotFound etc during this run and in subsequent runs the below mentioned Zip extension during verification. So why store these files in cache in the first place?  Or why not delete these files from cache during this run itself?

in-fke: Shouldn't it also be the case that HTTP request somewhat failed with an exception (e.g. broken connection)?
In this case the cache file should not be produced as well.
And only comparing "expectedBytes" would not work if the HTTP Response did not tell you the size to expect?

janakmulani: The PR is still under test. I am going to test the scenario you have pointed out. Thanks for pointing out.

sclassen: > Shouldn't it also be the case that HTTP request somewhat failed with an exception (e.g. broken connection)?

Yes, unfortunately our tests have shown that in some scenarios the inputstream just ends without giving an exception. Thus making it impossible to detect a corrupted file.
That is also way we putin the safety net with the downloaded bytes.

in-fke: > Yes, unfortunately our tests have shown that in some scenarios the inputstream just ends without giving an exception
But then the exception is probably "caught" somewhere in the inner workings.
What HTTP Client are you using and maybe you can point me to the relevant source file where the actual download takes place?

janakmulani: sun.net.www.protocol.https.HttpsURLConnectionImpl is used.
Note that httpConnection.setConnectTimeout and httpConnection.setReadTimeout are not set => infinite wait. when you break the connection, the download just waits.
All code is in BaseResourceDownloader. The `download` method is called to download jars. Currently the byte count check is done in `tryDownloading(DownloadDetails)` method. Perhaps in `tryDownloading(URL)` in the final catch statement we could add code to delete the cached file something like below:
`
} catch (Exception ex) {
            final URL resourceHref = resource.getLocation();
            final VersionId version = getVersion(downloadDetails.downloadFrom, downloadDetails.version);
            final File cacheFile = Cache.getCacheFile(resourceHref, version);
            if (cacheFile != null && cacheFile.exists()) {
                cacheFile.delete();
            }
            LOG.debug("Exception while downloading resource {} from {} - message: {} cause: {} ", resource, downloadFrom, ex.getMessage(), ex.getCause());
            throw ex;
        }
`

in-fke: Sounds like a plan (deleting the cache file in case of exception).

PR #804, opened 2021-05-27
create parent dir early and fail gracefully if parent is read only

sclassen:


common/src/main/java/net/adoptopenjdk/icedteaweb/lockingfile/LockableFile.java
```diff
@@ -90,7 +90,8 @@ private LockableFile(final File file) {
         this.file = file;
         try {
             if (!file.exists()) {
-                if (!file.createNewFile()) {
+                createParentDirIfMissing(file);
+                if (canWriteParent(file) && !file.createNewFile()) {
```
janakmulani, line 94: should we log debug the conditions that either parent does not have  write permission or createNewFile failed?

janakmulani: Other than above looks ok.

PR #810, opened 2021-06-14
Http connect and read timeout

janakmulani:


core/src/main/java/net/sourceforge/jnlp/config/Defaults.java
```diff
@@ -565,6 +565,18 @@
                     ConfigurationConstants.KEY_JVM_ARGS_WHITELIST,
                     null,
                     null
+            ),
+
+            Setting.createDefault(
+                    ConfigurationConstants.KEY_HTTPCONNECTION_CONNECT_TIMEOUT,
+                    String.valueOf(10000),
+                    null
```
sclassen, line 573: Can you use `ValidatorFactory.createRangedIntegerValidator(0, Integer.MAX_VALUE)`

core/src/main/java/net/sourceforge/jnlp/config/Defaults.java
```diff
@@ -565,6 +565,18 @@
...
+            ),
+
+            Setting.createDefault(
+                    ConfigurationConstants.KEY_HTTPCONNECTION_READ_TIMEOUT,
+                    String.valueOf(10000),
+                    null
```
sclassen, line 579: same validator as for CONNECT_TIMEOUT

core/src/main/java/net/adoptopenjdk/icedteaweb/resources/downloader/BaseResourceDownloader.java
```diff
@@ -152,7 +154,20 @@ private Resource tryDownloading(final URL downloadFrom) throws IOException {
...
+    }
+
+    private int getTimeoutValue(final String key) {
+        int timeout = 0;
+        final String value = JNLPRuntime.getConfiguration().getProperty(key);
+        if (value != null && value.trim().length() != 0) {
+            try {
+                timeout = Integer.valueOf(value);
+            } catch (NumberFormatException e) {
+                LOG.error(IcedTeaWebConstants.DEFAULT_ERROR_MESSAGE, e);
```
sclassen, line 167: Please use something else than `DEFAULT_ERROR_MESSAGE`, maybe
`Log.error("Could not pares {} with value {} - reason {}", key, value, e.getMessage());`

sclassen (changes requested):

core/src/main/java/net/sourceforge/jnlp/config/Defaults.java
janakmulani: done

core/src/main/java/net/sourceforge/jnlp/config/Defaults.java
janakmulani: done

core/src/main/java/net/adoptopenjdk/icedteaweb/resources/downloader/BaseResourceDownloader.java
janakmulani: done

PR #813, opened 2021-07-26
encode the location of the policy URL

sclassen:


core/src/main/java/net/sourceforge/jnlp/runtime/JNLPPolicy.java
```diff
@@ -235,13 +234,14 @@ private Policy getPolicyFromUrl(String policyLocation) {
             try {
                 final URI policyUri;
                 if (policyLocation.startsWith("file://")) {
-                    policyUri = new URL(policyLocation).toURI();
+                    final String encoded = URLEncoder.encode(policyLocation, UTF_8.name());
+                    policyUri = new URL(encoded).toURI();
                 } else {
                     policyUri = new URI(policyLocation.replace("\\", "/"));
```
janakmulani, line 240: should we encode here too?
In general we allow specification of urls in the deployment properties file. Should we encode all Urls read by OWS from the deployment properties file?

sclassen: Was thinking the same but I currently do not have the time to find all usages of URLs in the settings.
For now we do not have any other bug report so I would not change any more code.

core/src/main/java/net/sourceforge/jnlp/runtime/JNLPPolicy.java
```diff
@@ -235,13 +234,14 @@ private Policy getPolicyFromUrl(String policyLocation) {
             try {
                 final URI policyUri;
                 if (policyLocation.startsWith("file://")) {
-                    policyUri = new URL(policyLocation).toURI();
+                    final String encoded = URLEncoder.encode(policyLocation, UTF_8.name());
+                    policyUri = new URL(encoded).toURI();
```
janakmulani, line 238: This needed to be tested. For policyLocation file:///c:/temp/ows/AB/myjava.policy  new URL() throws exception java.net.MalformedURLException: no protocol: file%3A%2F%2F%2Fc%3A%2Ftemp%2Fows%2FAB%2Fmyjava.policy.

sclassen: did fix the unit test to detect malformed URL and also updated the implementation

PR #821, opened 2021-08-12
Append only cache file

sclassen:


sclassen:
@hendrikebbers Janak and I tested this. It is good to go

PR #831, opened 2021-10-19
set allowRedirect = true by default

bourgesl:
I propose to allow redirect by default.
How should the -allowRedirect flag be adapted ?


bourgesl:
Any feedback ?
Does any security reason requires allowRedirect = false ?

Approved by default ?

bourgesl:
Any feedback on a potential security issue ?

sclassen:
Sorry the release had priority for us.

We discussed this change in the team and we decided to leave the default as it is.
Main reasons for the decision were:
- changing the default could have unwanted effects on many applications.
- redirecting by default could open an attack vector.

Regards
Stephan

PR #841, opened 2021-11-16
Fix collision of shortcut icon names

janakmulani:


core/src/main/java/net/sourceforge/jnlp/util/XDesktopEntry.java
```diff
@@ -349,16 +349,20 @@ private void cacheIcon() throws IOException, NonFileProtocolException {
...
-            target = new File(PathsAndFiles.ICONS_DIR.getFile(), targetName);
+            target = new File(PathsAndFiles.ICONS_DIR.getFile(), getPrefixedName(targetName));
             Files.copy(cacheFile.toPath(), target.toPath(), StandardCopyOption.REPLACE_EXISTING);
         }
         this.iconLocation = target.getAbsolutePath();
         LOG.debug("Cached desktop shortcut icon: " + target + " ,  With source from: " + cacheFile.getAbsolutePath());
     }
 
+    private String getPrefixedName(final String targetName) {
+        return file.getSourceLocation().toString().hashCode()+"-"+targetName;
```
janakmulani, line 363: Should we use "-"? Also a possibility of hashCode being negative. May be we take absolute value or remove the minus sign?

PR #854, opened 2022-02-10
Change logging for lockable file

sclassen:


common/src/main/java/net/adoptopenjdk/icedteaweb/lockingfile/LockableFile.java
```diff
@@ -342,26 +342,25 @@ public boolean tryLock() throws IOException {
...
         }
 
         @Override
         public void unlock() throws IOException {
-            if (!lockFile.delete()) {
-                if (lockFile.exists()) {
-                    logger.error("Failed to delete lock file {}", lockFile);
-                }
+            if (lockFile.delete()) {
+                logger.error("Deleted lock file {}", lockFile);
```
janakmulani, line 353: Shouldn't this be logger.debug?

sclassen: absolutely

PR #874, opened 2022-05-31
add nashorn based pac evaluator

dimitryc:
Please review the patch that introduces an alternative Nashorn-based implementation for PAC (proxy-auto-config) functionality.

Switching to Nashorn will help to fix various compatibilities issues (that worked fine with Oracle JRE). E.g. an application may bundle Rhino itself and different versions of Rhino may conflict with each other.

The proposed patch doesn't change the behaviour if --with-rhino option is explicitly passed to configure script. It should be enough not to pass --with-rhino option to activate Nashorn-based implementation.

As per [1], the implementation doesn't use "eval(String)" method as the PAC scripts will treated as untrusted. Instead, the needed set of permissions is granted to the PAC scripts via Policy.getPermissions().

[1] https://wiki.openjdk.java.net/display/Nashorn/Nashorn+script+security+permissions

### DIFF
--- a/netx/net/sourceforge/jnlp/runtime/JNLPPolicy.java
+++ b/netx/net/sourceforge/jnlp/runtime/JNLPPolicy.java
@@ -17,13 +17,14 @@
 package net.sourceforge.jnlp.runtime;
 
 import java.io.File;
-import java.net.URI;
-import java.net.URISyntaxException;
+import java.net.*;
 import java.security.*;
 import java.util.Enumeration;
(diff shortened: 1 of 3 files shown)


ebourg:
Nashorn is no longer included in the JRE and the project seems abandonned. I'd rather embed and relocate the rhino dependency to solve the conflict issue.

sclassen:
Will close this as Nashorn is no longer included in the JREs

PR #898, opened 2023-04-23
Use same naming scheme like bash-completion

robert-scheck:
As of writing, all bash-completion files installed by IcedTea-Web are named e.g. `/usr/share/bash-completion/completions/javaws.bash` while bash-completion itself uses a naming scheme without `.bash`, thus e.g. `/usr/share/bash-completion/completions/javaws` would conform upstream's naming scheme.

Example file lists from downstreams packaging bash-completion:
- Fedora: https://packages.fedoraproject.org/pkgs/bash-completion/bash-completion/fedora-rawhide.html#files
- Debian: https://packages.debian.org/sid/all/bash-completion/filelist


robert-scheck:
Ping?

PR #908, opened 2023-05-30
Fix javadoc error related to @param in TimedHashMap.java (1.8)

robert-scheck:
When building using OpenJDK Java 17, I run into the following javadoc error related to `@param` in `TimedHashMap.java` which should be fixed:

```
IcedTea-Web-icedtea-web-1.8.8/netx/net/sourceforge/jnlp/util/TimedHashMap.java:60: error: invalid use of @param
 * @param K The key type
   ^
IcedTea-Web-icedtea-web-1.8.8/netx/net/sourceforge/jnlp/util/TimedHashMap.java:61: error: invalid use of @param
 * @param V The Object type
   ^
```


robert-scheck:
Ping?

PR #929, opened 2023-12-05
use reflection to call method with changed signature in Java 21

sclassen:


janakmulani:
Tested with https://spdf.gsfc.nasa.gov/skteditor/skteditor.jnlp

PR #958, opened 2024-10-17
Enable basic HTTP cookie support in Boot.java

lso-du-sud-ouest:
Enable java.net built-in support of HTTP cookies. HTTP cookies will be stored in-memory until the java process dies.

### DIFF
--- a/core/src/main/java/net/sourceforge/jnlp/runtime/Boot.java
+++ b/core/src/main/java/net/sourceforge/jnlp/runtime/Boot.java
@@ -46,6 +46,9 @@ import javax.jnlp.BasicService;
 import javax.swing.UIManager;
 import java.io.File;
 import java.net.URL;
+import java.net.CookieManager;
+import java.net.CookiePolicy;
+import java.net.CookieHandler;
 import java.security.AccessController;
 import java.security.PrivilegedAction;
 import java.util.Arrays;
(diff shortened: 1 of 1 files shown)


janakmulani:
Thanks for this PR. We will have look.

janakmulani:
@lso-du-sud-ouest You are setting Cookie Manager (CM) in Boot. OWS has two phases. Phase I is where OWS javaws is launched with the bundled jre (comes with installer). Phase I downloads the jnlp (if not already in cache). The Phase I then launches Phase II with the jre that best matches the one specified in the jnlp file. Phase II downloads jar files specified in the jnlp (if not already in cache). Boot is called in both the phases => CM will be set in each phase => cookies that were set while downloading the jnlp in Phase I will not be available to Phase II and the application. The cookies set during downloading of jars in Phase II will be available to the application started by jnlp. However, if the application is setting its own CM then the cookies from Phase II will not be available to the application.

lso-du-sud-ouest:
@janakmulani I fully agree with your description of Phases I and II. Let me describe how JWS applications are used in my company.

All JWS application resources (jnlp files, jar files, ...) are hosted on servers which require a user authentication. Every time, a user wants to launch one of our JWS application, he has to download the jnlp file from a secured web site and then he has to double-click the downloaded jnlp file.

Regarding the way OWS behaves in such context, I understand that Phase I doesn't download any file. I'm quite sure about it because the download of the 1st jar file triggers a basic authentication form which is displayed only once (for a successful authentication). All the other resources are automatically downloaded thanks to the authentication/session cookies received with the 1st download.

I know that other ways of launching JWS applications are possible, but without http cookie support, OWS cannot be used in my company context. http cookies are well supported by Oracle JRE, but we want to replace it with OpenJDK/OWS. And finally for us not being able to re-use Phase I cookies is not an issue.

janakmulani:
@lso-du-sud-ouest Thanks for clarifying about Phase I cookies. It should also be clear that if the application itself is creating a CookieManager then the cookies collected by OWS while downloading jars will be lost. If these two aspects are understood then we can proceed with your contribution.

PR #967, opened 2025-02-18
Add Property to enabled/disable the securityManager

WeierE1:
Add a property to disable the SecurityManager. 
With this it is possible to disable the SecurityManager via Property.


janakmulani:
@WeierE1 Thanks for this PR. We already have this on our agenda, i.e., a property in as an additional alternative to -nosecurity cmdline option. We will do this in the next release.

WeierE1:
Sounds great. Do you now when the next release will be build?

janakmulani:
See https://github.com/AdoptOpenJDK/IcedTea-Web/pull/977

PR #970, opened 2025-05-02
Extend JAVADOC_OPTS for Java 21+ (1.8)

robert-scheck:
https://salsa.debian.org/java-team/icedtea-web/-/blob/master/debian/patches/jdk-21-autoconf.patch is a prerequisite, allows removing `--ignore-source-errors` from `JAVADOC_OPTS` being introduced with commit be989488bb60c23b8ea2b5e95282d661d25a4077.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -287,7 +287,15 @@ if ENABLE_DOCS
 JAVADOC_OPTS=-use -keywords -encoding UTF-8 -splitIndex \
  -bottom '<font size="-1"> <a href="http://icedtea.classpath.org/bugzilla">Submit a bug or feature</a></font>'
 if HAVE_JAVA9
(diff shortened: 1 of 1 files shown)


janakmulani:
@robert-scheck As can be seen in the method getPredefinedJavaModulesVMArgumentsMap() of  [JvmUtils.java](https://github.com/AdoptOpenJDK/IcedTea-Web/blob/master/core/src/main/java/net/adoptopenjdk/icedteaweb/jvm/JvmUtils.java), there are additional jvm args required when running with Java 9+. Please add the missing arguments and resubmit the PR.

lherschi:
On the one hand the patch for acinclude.m4 is missing in the commit and on the other hand this breaks the build with Java 11. The 1.8 branch cannot be built with Java 21 anyway, so why the pull request here at all?

PR #971, opened 2025-05-02
Prepend -Djava.security.manager as workaround for Java 21 (1.8)

robert-scheck:
The Java security manager is deprecated since Java 17 and disabled by default since at least Java 21 (and subject to removal in a future release). For the time being enabling the security manager (as a workaround) leads to a working IcedTea-Web using Java 21.

Inspired by https://salsa.debian.org/java-team/icedtea-web/-/blob/master/debian/patches/javaws_change_java_policy.diff

### DIFF
--- a/rust-launcher/src/main.rs
+++ b/rust-launcher/src/main.rs
@@ -172,6 +172,7 @@ fn compose_arguments(java_dir: &std::path::PathBuf, original_args: &std::vec::Ve
     all_args.push(cp);
     all_args.push(bin_name);
     all_args.push(bin_location);
+    all_args.push(String::from("-Djava.security.manager"));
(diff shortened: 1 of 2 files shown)


janakmulani:
@robert-scheck In [OpenWebStart ](https://github.com/karakun/OpenWebStart) we set a JVM arg -Djava.security.manager=allow. Are you sure just -Djava.security.manager is sufficient?

robert-scheck:
@janakmulani, yes, it works for me for Fedora 41 and 42 as well as on RHEL 9 and 10 with at least OpenJDK Java 17 and 21. And I guess it also works for Debian, because this is where I took this from. I am happy to change this if this pull request will be merged afterwards.

Issue #972, opened 2025-05-06
Application fails to start when user's home folder contain certain characters


fthevenet:
We recently received a report where an application launched via a JNLP file would failed to start for one user while it worked fine for others when using Icedtea-web 1.8 with OpenJDK 8 on Windows, while it was working for everyone when they used Oracle WebStart.
After investigation, the issue turned out to be that user for whom the application wouldn't start had an apostrophe in their home folder (e.g. `C:\Users\Foo_O'Bar`), which is a legal path on Windows, but that ITW code for crafting new paths for downloaded asset would incorrectly sanitize as `C:\Users\Foo_O_Bar`.
The application would then fail as it attempted to access content in  `C:\Users\Foo_O_Bar\...` instead of  `C:\Users\Foo_O`Bar\...` 

I would like to propose a fix and I have opened PRs for both the `1.8` and `master` branches.

janakmulani:
Hi

`C:\IcedTeaWeb\WebStart\bin>javaws \owsfiles\Foo_O'Bar\AccessibleScrollDemo8.jnlp` throws `java.io.IOException: Cant create directory C:\Users\Foo_O_Bar\.cache\icedtea-web\cache\1`.

However [OpenWebStart](https://github.com/karakun/OpenWebStart) works `C:\OWS_1_11_1>javaws \owsfiles\Foo_O'Bar\AccessibleScrollDemo8.jnlp`

fthevenet:
Hello,
I'm afraid I don't quite follow you; are you confirming that the issue exists in Icetea-web, but that it doesn't happen in OpenWebStart?
I am not entirely knowledgeable about the relationship between ITW and OWS; does OWS consumes ITW as a dependency, or is it a downstream fork?
At any rate, this is the tracker for the ITW project, so it seems correct to report and propose a fix for this here, regardless of whether OWS is affected or not,, is it not?

Thanks

janakmulani:
I confirm that the issue exists in ITW but it does not happen in OWS.

OWS does depend on ITW https://github.com/AdoptOpenJDK/IcedTea-Web

Sure this is a tracker for ITW and you are right to propose a fix here.

fthevenet:
Thanks for the precisions, @janakmulani.
I'm looking forward to a review for #973 

Thanks.

PR #973, opened 2025-05-06
Do not apply sanitization to root path when generating paths for downloaded assets in the cache folder

fthevenet:
This PR aims to address issue #972 

Application launched via a JNLP file will failed to start for user wtih certain characters in their home folder (e.g. `C:\Users\Foo_O'Bar`), which is a legal path on Windows, but paths for downloaded asset are incorrectly sanitize as `C:\Users\Foo_O_Bar` and  the application then fails as it attempts to access content in `C:\Users\Foo_O_Bar\...` instead of `C:\Users\Foo_OBar...`

### DIFF
--- a/core/src/main/java/net/sourceforge/jnlp/cache/CacheUtil.java
+++ b/core/src/main/java/net/sourceforge/jnlp/cache/CacheUtil.java
@@ -147,9 +147,6 @@ public class CacheUtil {
 
         StringBuilder path = new StringBuilder();
 
-        path.append(root);
-        path.append(File.separatorChar);
-
         path.append(location.getProtocol());
         path.append(File.separatorChar);
         path.append(location.getHost());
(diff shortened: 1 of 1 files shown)


janakmulani:
@fthevenet we think removing the character "'" from INVALID_PATH list in [FileUtils.java](https://github.com/AdoptOpenJDK/IcedTea-Web/blob/master/common/src/main/java/net/adoptopenjdk/icedteaweb/io/FileUtils.java) should fix the issue of "Foo_O'Bar" as "'" is a valid character for filepath in Windows. Can you please try this fix and resubmit the PR?

fthevenet:
I don't think this is the right approach; even though it would indeed fix this particular occurrence.
For example, the same problem would arise with a path containing the character  `:` on Linux, which ,while legal on this file system, would be incorrectly sanitized and result in the same `FileNotFoundException`.

On the other hand, I think the `root` parameter used by the method should always contain a path that is legal for the runtime platform, since it is derived from system properties or supplied by the user via a property, so applying the sanitation to this part of the path gains us nothing.

PR #994, opened 2026-02-17
Changes to support launch on java 24 or higher. This is based on the changes of PR #969

akuhtz:
Changes to support launch on java 24 or higher. This is based on the changes of PR #969


core/src/main/java/net/sourceforge/jnlp/runtime/JNLPRuntime.java
```diff
@@ -603,15 +619,20 @@ public static void setExitClass(Class<?> exitClass) {
...
     /**
      * @return the current Application, or null if none can be
      * determined.
      */
     public static ApplicationInstance getApplication() {
-        return security.getApplication();
+        if (security != null) {
+            return security.getApplication();
+        }
+        return null;
```
janakmulani, line 635: This is problematic for Jnlp Services.

janakmulani:
PR https://github.com/AdoptOpenJDK/IcedTea-Web/pull/969 has some issues. Moreover, the changes related to the  SecurityManager also have some issues. Will create a new branch and new PR after testing various scenarios. This will need time and effort, hence won't be part of the next release (due likely in June).

akuhtz:
@janakmulani Please add reference to this PR with your new solution.

guykatz:
very critical for me also. @akuhtz since you opened a PR to solve this, and assuming I dont really care too much about the things that @janakmulani stated as problematic,  is there a bypass i can use if I need tumerin 25 support right now?
thanks!

PR #1000, opened 2026-07-28
Update github workflow

lherschi:


lherschi:
We need to migrate away from CentOS 7 as a container. I’m therefore closing this PR.

PR #1002, opened 2026-08-17
fix: open Control Panel Simple editor without --add-exports

martinhickson:
## Summary
- Bare `java -cp` / `groovy -cp` Control Panel was skipping the in-process Simple editor and forking `policyeditor`. That is not a PolicyEditor window in the Control Panel JVM.
- Export `sun.security.provider` to the unnamed module at click time so `PolicyEditor` can be constructed in-process, then show the window.
- If that export is impossible, still launch this install's `bin/policyeditor` (or `java -cp` with modular flags) with null-device stdio so a full pipe cannot stall the child.

## Test plan
- [ ] `icedtea-web-settings` → Policy Settings → Simple editor opens a PolicyEditor window (Temurin 21)
- [ ] `groovy -cp icedtea-web-uber.jar` `new ControlPanel(config)` **without** `--add-exports` → Simple editor opens a PolicyEditor **window** (`window=true`), not only a child process
- [ ] Same path **with** `--add-exports` still opens in-process (`window=true process=false`)
- [ ] Standalone `policyeditor -file …` still opens a window
- [ ] `mvn -pl icedtea-web test -Dtest=ItwLauncherPathsTest`


martinhickson:
Opened against the wrong repository; closing.

